PR #2, opened 2013-12-24
performance: use Linux extended syscalls and flags

gstrauss:
reduce syscalls on Linux using extended syscalls and flags,
e.g. accept4(), pipe2(), O_CLOEXEC, SOCK_CLOEXEC, SOCK_NONBLOCK


gstrauss:
rebased against tip of master, resolved conflicts, and force-pushed updated patch branch

PR #3, opened 2014-06-10
support proxy-protocol on listening sockets

russor:
Support for the proxy-protocol http://haproxy.1wt.eu/download/1.5/doc/proxy-protocol.txt

This is tested (applied to 1.4.35) on FreeBSD. Basically lets us put our lighttpds behind a proxy without losing the client IP.

if proxy.expect_header is set in socket context, the header is read and processed; if proxy.expect_header & 2; then the schema is set to https (proxy is doing TLS termination, but lighttpd needs to know)

There's a small memory cost to have the additional state, so I put it all in ifdefs, with a configure option.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -606,6 +606,18 @@ if test x$mmap = xtrue; then
   AC_DEFINE(ENABLE_MMAP, [1], [Use mmap if available])
 fi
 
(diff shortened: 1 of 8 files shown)


stbuehler:
not merged - closing for now. please create a new pull request with an updated patch if you are still interested.

PR #5, opened 2014-10-20
log

demiaowu:
/**
 * From whole sources of the lighttpd1.4, we can find that the second and third parameter in the
 * function log_error_write and log_error_multiline_buffer are __FILE__, __LINE__ respectively.
 * So we can define two macros: log_error_write_caf,log_error_write_multiline_buffer_caf instead
 * Apparently, They avoid a lot of repeated operations. Certainly, It is not only compatible, but also
 * convenient and fast for the future's use;
 * 	"convenient and fast"  logogram: caf
 * 	So I Use of 'caf' as a suffix of the macros.
 */


stbuehler:
No.

* completely screwed git history
* C comments are not commit messages
* the logging could be simplified, yes (but I don't see much value in changing it now). "_caf" is not not a suitable suffix, and macros should always be upper case.
* the "perfect" key-value handling looks crazy and broken. no way. And don't try to "optimize" it unless you can show it is a bottleneck, and it actually solves something.

PR #6, opened 2015-07-07
Adds support for m3u8 files

automatical:
Adds support for an m3u8 mimetype.

I think it's open to discussion whether lighttpd can stick to the lower case convention, or use application/x-mpegURL instead.


stbuehler:
Fixed in 8db141a1b30e5808679e1f341c6ca914be91bed8 by updating create-mime.conf.pl and rerunning it - thanks for pointing out that something was missing :)

PR #8, opened 2015-07-19
Correcting config_parse_file()

speedyb0y:
The compiler gave me this error:

src/configfile.c:1232:8: error: 's.size' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     if (s.size == 0) /* the file was empty, nothing to parse */

- I think this way the generated code will be smaller - it's the stream_open() that should check the stat->size anyway...
- Also, we are calling stream_close() even when stream_open() did fail...

Sorry for my english, for my weird coding style and for my stupidness if I'm doing anything wrong here...

### DIFF
--- a/src/configfile.c
+++ b/src/configfile.c
@@ -994,23 +994,31 @@ int config_parse_file(server *srv, config_t *context, const char *fn) {
 		filename = buffer_init_buffer(context->basedir);
 		buffer_append_string(filename, fn);
 	}
+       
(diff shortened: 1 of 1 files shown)


stbuehler:
Thanks for reporting the read of an uninitialized variable!

I fixed it by making sure no error gets returned for empty files - see def17b2925f5e8db74a7dbd93448726adac41e0c

PR #9, opened 2015-08-20
Import explicit_bzero() and use it 

loganaden:
Problem: Compilers may optimize away calls to memset due to their optimization techniques. As shown in the snippet below, memset may be optimized away as the final variable is no longer used. To prevent this, we use OpenBSD's explicit_bzero() API which is designed not to be optimized. It's stable enough that it ships in LibreSSL. Since final is sensitive, it's safer to use explicit_bzero.

-----
    /*
     * Don't leave anything around in vm they could use.
     */
    memset(final, '0', sizeof(final));

	/* FIXME
	 */
#define apr_cpystrn strncpy
    apr_cpystrn(result, passwd, nbytes - 1);
----

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -74,7 +74,8 @@ common_src=buffer.c log.c \
 	network_write.c network_linux_sendfile.c \
 	network_freebsd_sendfile.c network_writev.c \
 	network_solaris_sendfilev.c network_openssl.c \
-	splaytree.c status_counter.c
+	splaytree.c status_counter.c \
+	explicit_bzero.c
 
 src = server.c response.c connections.c network.c \
 	configfile.c configparser.c request.c proc_open.c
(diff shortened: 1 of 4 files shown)


stbuehler:
The ``explicit_bzero`` implementation looks like an unreliable hack to me

loganaden:
It is a hack to deal with "smart" compilers :)

It has been adapted by other projects as well:
Unbound and NSD use it:
https://github.com/jedisct1/unbound/blob/master/compat/explicit_bzero.c
FreeBSD has imported it:
http://www.unix.com/man-page/freebsd/3/explicit_bzero/

Linux has adapted explicit_bzero and named their implement memzero_explicit:
https://lkml.org/lkml/2014/8/25/497

stbuehler:
linux uses a different implementation (varying across compilers, so not portable in general unless you do the work for all compilers).

Maybe we can add the `volatile` hacks from https://www.mail-archive.com/openssl-dev@openssl.org/msg34134.html

I think the solution should something like:
- maybe test whether `libc` has `memset_s` - if yes, use it (i don't know of any implementation yet...)
- test whether `libc` has `explicit_bzero` - if yes, use it
- otherwise run own implementation:
  - `volatile` hacks from https://www.mail-archive.com/openssl-dev@openssl.org/msg34134.html
  - test whether compiler supports `__attribute__((weak))`; if yes, add and use weak hook

loganaden:
Here's the tentative diff:

http://elandsys.com/~logan/lighttpd.diff

stbuehler:
Fixed in d7be04beb5757b00a1b25ab355693db13973e9a5

PR #10, opened 2015-09-18
Fix allocation failure of pParser

ya1gaurav:
Apply this fix to check allocation failure.


stbuehler:
Applied in 37bdb250a4942a7c12382e6fda072abec50bae8a, thanks

PR #11, opened 2015-09-26
check ssiexprparserAlloc() result with force assert

ya1gaurav:
In case allocation fails, pParser will be Null & it is derefrenced below.


stbuehler:
Applied in 5c5f67a5c7bbeca6a4b8eb92e43752499efea53a - thanks!

PR #12, opened 2015-09-26
check allocation failure at many places.

ya1gaurav:
Allocation failure can derefrence Null pointer at later stage, so added force_assert check.


stbuehler:
Applied in 5c5f67a5c7bbeca6a4b8eb92e43752499efea53a - thanks!

PR #13, opened 2015-12-09
[mod_csrf] module to aid against csrf attacks

jin-eld:
The module implements a cookie-to-header protection mechanism by sending
out a token as a cookie and by expecting the data of the cookie in a
header of the incoming request.

The validation of the token only takes place for configured "protected" URLs,
so an "entry" to a page (i.e. index.html) can be unprotected and thus be
used for obtaining a cookie.

The token is based on an sha1 hmac where the input data are the
timestamp and the username (if available). The secret key for the hmac
routine is generated by the SSL random function and is stored in memory.

The expiration time of the token is configurable, the token will be
updated and renewed before the expiration time elapses.

The module requires OpenSSL functions and will only be enabled if
lighttpd is compiled with OpenSSL support.

Configuration options are:

Name of the cookie that will be sent in outgoing responses:
csrf.send-cookie-name = "csrf";

Name of the header that will contain the token value in incoming requests:
csrf.receive-header-name = "X-Csrf-Token";

Cookie "Domain" setting, not added to the cookie if unset:
csrf.cookie-domain = "";

Cookie lifetime in seconds:
csrf.cookie-lifetime = 600;

Conditions can be used to activate protection for certain URLs:

$HTTP["url"] =~ "^/(someurl|cgi-bin)/(.+)" {
    csrf.activate = "enable"
}

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,6 +8,8 @@ AC_CONFIG_MACRO_DIR([m4])
 
 AC_CANONICAL_TARGET
 
+MOD_CSRF=0
+
 dnl HOW FUCKING BROKEN IS AUTOMAKE...
 dnl serial-tests is not recognized before 1.12, and required for serial tests with 1.13, sane version checks "impossible"
 dnl if you don't use the default (symlinked) automake on your box, set AUTOMAKE to the path you're using
@@ -354,10 +356,13 @@ if test "x$use_openssl" = "xyes"; then
     OLDLIBS="$LIBS"
     AC_CHECK_LIB(crypto, BIO_f_base64, [
       AC_CHECK_LIB(ssl, SSL_new, [ SSL_LIB="-lssl -lcrypto"
-				 AC_DEFINE(HAVE_LIBSSL, [], [Have libssl]) ], [], [ -lcrypto "$DL_LIB" ])
+				 AC_DEFINE(HAVE_LIBSSL, [], [Have libssl])
+                 MOD_CSRF=1 ], [], [ -lcrypto "$DL_LIB" ])
     ], [], [])
     LIBS="$OLDLIBS"
     AC_SUBST(SSL_LIB)
+else
+    MOD_CSRF=0
 fi
 
 AC_MSG_CHECKING(for perl regular expressions support)
@@ -564,6 +569,9 @@ AC_SEARCH_LIBS([crypt_r],[crypt],[
     ])
 
     CRYPT_LIB=$LIBS
+  ],
+  [
+    MOD_CSRF=0
   ])
 ])
 LIBS=$save_LIBS
@@ -712,6 +720,11 @@ dnl build version-id
 LIGHTTPD_VERSION_ID=`echo $PACKAGE_VERSION | $AWK -F '.' '{print "(" $1 " << 16 | " $2 " << 8 | " $3 ")"}'`
 AC_DEFINE_UNQUOTED([LIGHTTPD_VERSION_ID], [$LIGHTTPD_VERSION_ID], [lighttpd-version-id])
 
+if test "x$MOD_CSRF" = "x1"; then
+    CSRF_CRYPT_LIB="$CRYPT_LIB"
+    AC_SUBST(CSRF_CRYPT_LIB)
+fi
+
 AC_CONFIG_FILES([Makefile src/Makefile
 		 doc/config/conf.d/Makefile \
 		 doc/config/vhosts.d/Makefile \
@@ -844,6 +857,12 @@ else
 	disable_feature="$disable_feature $features"
 fi
 
+plugin="mod_csrf"
+if test "x$MOD_CSRF" = "x1"; then
+	do_build="$do_build $plugin"
+else
+	no_build="$no_build $plugin"
+fi
 
 dnl output
 
--- a/doc/config/conf.d/Makefile.am
+++ b/doc/config/conf.d/Makefile.am
@@ -3,6 +3,7 @@ EXTRA_DIST=access_log.conf \
 	cgi.conf \
 	cml.conf \
 	compress.conf \
+	csrf.conf \
 	debug.conf \
 	dirlisting.conf \
 	evhost.conf \
--- a/doc/config/conf.d/csrf.conf
+++ b/doc/config/conf.d/csrf.conf
@@ -1,0 +1,39 @@
+#######################################################################
+##
+##  CSRF Protection Module
+## ----------------
+##
+server.modules += ( "mod_csrf" )
+
+##
+## Use conditions to activate protection for certain URLs:
+#$HTTP["url"] =~ "^/(someurl|cgi-bin)/(.+)" {
+#    csrf.protection = "enable"
+#}
+
+##
+## Cookie lifetime in seconds
+## Default: 600 seconds (10 minutes)
+##
+#csrf.cookie-lifetime = 600;
+
+##
+## Cookie domain, not set if empty
+## Default: empty
+#csrf.cookie-domain = "";
+
+##
+## Name of the cookie that is being sent out
+## Default: "csrf"
+##
+#csrf.send-cookie-name = "csrf";
+
+##
+## Name of the http header that is being expected in incoming requests
+## Default: "X-Csrf-Token"
+##
+#csrf.receive-header-name = "X-Csrf-Token";
+
+##
+#######################################################################
+
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -187,6 +187,11 @@ mod_usertrack_la_SOURCES = mod_usertrack.c
 mod_usertrack_la_LDFLAGS = -module -export-dynamic -avoid-version
 mod_usertrack_la_LIBADD = $(common_libadd)
 
+lib_LTLIBRARIES += mod_csrf.la
+mod_csrf_la_SOURCES = mod_csrf.c
+mod_csrf_la_LDFLAGS = -module -export-dynamic -avoid-version
+mod_csrf_la_LIBADD = $(CSRF_CRYPT_LIB) $(SSL_LIB) $(common_libadd)
+
 lib_LTLIBRARIES += mod_proxy.la
 mod_proxy_la_SOURCES = mod_proxy.c
 mod_proxy_la_LDFLAGS = -module -export-dynamic -avoid-version
--- a/src/mod_csrf.c
+++ b/src/mod_csrf.c
@@ -1,0 +1,476 @@
+#include "base.h"
+#include "log.h"
+#include "buffer.h"
+
+#include "plugin.h"
+
+#include <ctype.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if defined (USE_OPENSSL)
+
+#include <openssl/rand.h>
+#include <openssl/hmac.h>
+
+#define TOKEN_COOKIE_LIFETIME       10*60   /* seconds (10 minutes) */
+#define SECRET_SIZE_BYTES           20
+/* NOTE: if you decide to change the algo or to make it configurable,
+ * make sure not to forget about the TOKEN_SIZE_BYTES define */
+#define TOKEN_SIZE_BYTES            20      /* size of token in bytes, sha1 */
+#define TIMESTAMP_SIZE_STRING       20      /* max length of the timestamp */
+#define TIMESTAMP_SIZE_STRINGZ      TIMESTAMP_SIZE_STRING + 1
+#define TOKEN_SIZE_STRING           TOKEN_SIZE_BYTES * 2
+#define TOKEN_SIZE_STRINGZ          TOKEN_SIZE_STRING + 1
+#define TOKEN_SIZE_FULL_STRING      TIMESTAMP_SIZE_STRING + TOKEN_SIZE_STRING
+#define TOKEN_SIZE_FULL_STRINGZ     TOKEN_SIZE_FULL_STRING + 1
+/* plugin config for all request/connections */
+
+enum {
+    TOKEN_CHECK_OK,
+    TOKEN_CHECK_OK_RENEW,
+    TOKEN_CHECK_FAILED
+};
+
+typedef struct {
+    buffer *send_cookie_name;
+    buffer *cookie_domain;
+    unsigned int cookie_lifetime; /* life time of the cookie in seconds */
+    buffer *receive_header_name; /* max number of sessions, 0 == unlimited */
+    unsigned short protection;
+} plugin_config;
+
+typedef struct {
+    PLUGIN_DATA;
+
+    plugin_config **config_storage;
+    plugin_config conf;
+    unsigned char secret[SECRET_SIZE_BYTES];
+} plugin_data;
+
+static int mod_csrf_generate_token(plugin_data *p, const char *user,
+                                    time_t expires,
+                                    char *out, int out_len) {
+    struct tm gm;
+    unsigned char *digest;
+    unsigned int digest_len = 0;
+    buffer *data;
+
+    localtime_r(&expires, &gm);
+
+    data = buffer_init();
+    buffer_append_strftime(data, "%s", &gm);
+
+    if (user != NULL) {
+        buffer_append_string(data, user);
+    }
+
+    digest = HMAC(EVP_sha1(), p->secret, sizeof(p->secret),
+                      (unsigned char *)data->ptr, data->used,
+                      NULL, &digest_len);
+    if (digest == NULL) {
+        buffer_free(data);
+        return 0;
+    }
+
+    buffer_reset(data);
+    buffer_append_strftime(data, "%s", &gm);
+    buffer_append_string_encoded(data, (char *)digest,
+                                 digest_len, ENCODING_HEX);
+    memset(out, 0, out_len);
+    strncpy(out, data->ptr, out_len - 1);
+    buffer_free(data);
+    return 1;
+}
+
+int mod_csrf_check_token(server *srv, plugin_data *p,
+                                 const char* token, const char* user,
+                                 time_t *expires) {
+    char generated[TOKEN_SIZE_FULL_STRINGZ];
+    char sts[TIMESTAMP_SIZE_STRINGZ];
+    size_t ts_end;
+    struct tm tstm;
+    time_t ts;
+    int ret = TOKEN_CHECK_FAILED;
+    int renew = 0;
+    double diff;
+
+    if (!token) {
+        return TOKEN_CHECK_FAILED;
+    }
+
+    if (strlen(token) <= TOKEN_SIZE_STRING) {
+        log_error_write(srv, __FILE__, __LINE__, "s",
+                "token check failed: invalid token length");
+        return TOKEN_CHECK_FAILED;
+    }
+
+    ts_end = strlen(token) - TOKEN_SIZE_STRING;
+    if (ts_end > TIMESTAMP_SIZE_STRING) {
+        log_error_write(srv, __FILE__, __LINE__, "s",
+                "token check failed: invalid timestamp length");
+        return TOKEN_CHECK_FAILED;
+    }
+
+    memset(sts, 0, TIMESTAMP_SIZE_STRINGZ);
+    strncpy(sts, token, ts_end);
+    sts[TIMESTAMP_SIZE_STRINGZ - 1] = '\0';
+
+    if (strptime(sts, "%s", &tstm) == NULL) {
+        log_error_write(srv, __FILE__, __LINE__, "s",
+                "token check failed: could not parse timestamp");
+        return TOKEN_CHECK_FAILED;
+    }
+
+    ts = mktime(&tstm);
+    if (ts == -1) {
+        log_error_write(srv, __FILE__, __LINE__, "s",
+                "token check failed: invalid timestamp");
+        return TOKEN_CHECK_FAILED;
+    }
+
+    if (!mod_csrf_generate_token(p, user, ts,
+                                 generated, TOKEN_SIZE_FULL_STRINGZ)) {
+        log_error_write(srv, __FILE__, __LINE__, "s",
+                "token check failed: could not generate comparison token");
+        return TOKEN_CHECK_FAILED;
+    }
+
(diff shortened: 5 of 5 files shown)


stbuehler:
- I'm not happy about the Makefile stuff; it is different from what we do for the other modules I think, but still somehow strange... I think we usually build all modules, but build them "empty" if the dependencies are missing (not saying this is sane...).
- we are not using `<stdbool.h>` so far; I'd rather avoid it in case it breaks on some platforms
- we usually do URL matching using config conditionals. there are some exceptions (rewrite, redirect, ...), but I'd rather avoid adding to the list. Also I'd like to avoid the `<utlist.h>` dependency.
- the expiry isn't actually checked afaics
- subtracting from user supplied `ts` could trigger an overflow
- variable declarations must be at the beginning of a block

jin-eld:
> I'm not happy about the Makefile stuff; it is different from what we do for the other modules 
> I think, but still somehow strange... I think we usually build all modules, but build them "empty" if
> the dependencies are missing (not saying this is sane...).

I reworked the makefile so that it will build an empty csrf module now, similar to how it is done with the others. Actually I'd rather prefer it the other way around (why build empty modules at all?), but I understand it's probably a historical thing and you want to keep it consistent.

> we are not using <stdbool.h> so far; I'd rather avoid it in case it breaks on some platforms

Done.
    
> we usually do URL matching using config conditionals. there are some exceptions (rewrite,
> redirect, ...), but I'd rather avoid adding to the list. Also I'd like to avoid the <utlist.h> dependency.

Thanks for helping out on IRC, I got the idea, I hope it now works the way you meant it.

> the expiry isn't actually checked afaics

Good catch! :) Fixed.

> subtracting from user supplied ts could trigger an overflow

Moved the part that checks the cryptographic validity of the token up, before the ts calculations. So all calculations now take place with a ts that was generated by us. Imho this is probably enough, or am I still missing something?

> variable declarations must be at the beginning of a block

Fixed, hopefully I did not miss any.

jin-eld:
Added just a minor fix: I'm not using the pcre anymore, but configure was still disabling mod_csrf if pcre was not available. Removed that part.

jin-eld:
And one more minor updated: changed configuration parameter name from csrf.activate = "enable" to csrf.protection = "enable", I think it makes more sense that way, given that the module is always active when included, but only certain URLs are "protected".

stbuehler:
I don't see why the token is sent as cookie to the client; I'd use the same HTTP header for both directions. 

Also I'm not happy about parsing and writing timestamps as strings: for example if you sign time stamp "10" for user "test" the signature is also valid for time stamp "1" and user "0test".

I think we also already talked about using base64 instead of hex-encoding.

I now rewrote the whole thing (and changed other things as well): http://git.lighttpd.net/lighttpd/lighttpd-1.x.git/commit/?h=lighttpd-1.4.x-stbuehler-mod-csrf

Perhaps you could test it and write some unit tests for it :)

jin-eld:
Hi,

thanks a lot for finding time for it!

> I don't see why the token is sent as cookie to the client; I'd use the same HTTP header for both directions. 

No, no, please don' do that :) The thing is that you want to have some "entry" pages that can be accessed without the token, basically because they do not allow any "dangerous" actions. This is the way to obtain the token - just by loading "secure" static pages, like index.html in your browser will provide you with the token which you then need when you start calling cgi scripts or doing AJAX requests.

If you use headers both ways, then you can not obtain the token easily, you would have to make an extra AJAX request that would fail, then evaluate its response headers to get the token and then repeat your original request. As you can imagine - this would be quite a nightmare for web applications/clients...

The cookie to header token is a commonly used technique, so I'd really suggest staying with it. Please  refer to:
https://en.wikipedia.org/wiki/Cross-site_request_forgery#Cookie-to-Header_Token

I'm a bit swamped this week so I guess I'll only be able to do real testing and tryouts only in a week, but I will for sure try it and give you feedback, will also try to come up with some tests.

Thank you once again and please give the cookie approach another thought, if it stays header-only the module would be really difficult to use...

stbuehler:
The cookie approach works because the cookie is set on login; a login request is not cached.

In your case though you are using normal GET requests, which usually are cached; I have no idea what cookie headers from cached responses are doing, but whether they get re-evaluated or ignored - it doesn't do what you want.
=> For your scenario to work you'd have to disable all caching.

In other words: you really *really* need to handle CSRF-token timeouts in your Javascript framework. And you can try to get the first token from a GET request which already is in the cache, but be prepared it might be not valid anymore.

jin-eld:
> In other words: you really really need to handle CSRF-token timeouts in your Javascript framework. 
> And you can try to get the first token from a GET request which already is in the cache, but be prepared
>  it might be not valid anymore.

This is luckily not a problem - most frameworks provide a "beforerequest"/"beforeSend" callback which allows you to fetch the "newest" cookie and update the header of your next request. Since it can be done centrally in one place, it is very convenient to use.

jquery:


    $.ajaxPrefilter(function(options)
    {
            options.beforeSend = function(xhr)
            {
                var csrf = $.cookie("csrf-token");
                if (csrf)
                {
                    xhr.setRequestHeader('X-CSRF-Token', csrf);
                }
            }
    });

ExtJS:


      Ext.Ajax.on('beforerequest',
            function() {
                var csrf = Ext.util.Cookies.get("csrf-token");
                if (csrf && csrf.length > 0) {
                    Ext.Ajax.defaultHeaders = {
                        'X-Csrf-Token': csrf
                    };
                }
            }
        );

As mentioned earlier, I did not invent this technique, the "cookie to header" approach seems to be quite common and from my experience it is indeed easy to handle it on the client side.

I'll have time to finally take a closer look at the module this Monday, I will do a few tests and so on, but I'd really appreciate if we could stick to the commonly used "cookie to header" way of handling this.

stbuehler:
The generic rule regarding "Set-Cookie" headers seems to be that you should also explicitly disable caching in those responses (or use request methods which are not cached in the first place, like `POST`). I don't think this would be a good idea here, disabling caching on random requests just because they happen to need to update the CSRF token - which is why I don't want to use the cookie way, and would like to stick with the simple extra header solution.

> As mentioned earlier, I did not invent this technique, the "cookie to header" approach seems to be quite common and from my experience it is indeed easy to handle it on the client side.

This "common" technique was not designed to work without a POST login request, so please don't pull "but everyone else does it" arguments on me when you don't have technical arguments.

I thought we discussed this further in IRC but couldn't find anything in my logs.

jin-eld:
Yes we did, but the timeline got messed up a bit :) I posted my reply here before our conversation about caching on IRC, so the comment you replied to now is somewhat outdated, the caching problematic was not quite clear to me until you explained it on IRC.

I ended up using my version of the module, because it's easier to use for my scenario (I do not use caching on that platform and catching the token via a cookie is much easier). Nevertheless I'd be happy to see your CSRF version merged, because it's generally a good thing to have. I did try it and it worked as you designed it, it's just too much effort to adapt the existing UI and this is the only reason why I did not switch. Others, who maybe design new web UIs from scratch, might still find it useful.

gstrauss:
housekeeing: what is the status of this pull request?  should it be discarded in favor of @stbuehler version?  Has @stbuehler version been rebased against and tested with lighttpd git master?

jin-eld:
Well, I did not rebase it any more, because I did understnad that it's not going to be merged? We do use my module on our system, because it is easier to integrate in our environment and imho easier to use in general. @stbuehler did raise some points on why "cookie-to-header" might not always be the best way and for sure the concerns related to caching are also valid. At the same time I would not discard the the "cookie-to-header" approach, after all I think that it's "common practice" for pragmatic reasons,  it does simplify a couple of things on the UI side.

But anyway, I did a suggestion, we do use it in a production environment and it works well for us. If you are still interested, I will rebase, I can also adapt a few things if need be (i.e. make the tokens shorter as @stbuehler wanted). If not, then please close the PR. It's your decision in the end :)

gstrauss:
Thanks for the response.  Please hold off rebasing this specific pull request, since I do not want you to do work which might be discarded.  I guess the current status is: "waiting for further guidance".  Thanks.

danzimal:
I'm considering this module (Thanks for the solution!).
Do you think it's compatible with lighttpd 1.4.26?
We don't compile with OpENSSL (I'll find a suitable token generator :)

jin-eld:
From what I have seen so far the lighttpd module API changes frequently between releases, you would most likely have to port the module to 1.4.26 in order to make it compile, but I think you should be able to. I don't think that I am using anything "special" there, imho you'd just have to adapt to the API of 1.4.26 and it should work.

danzimal:
Thanks for this thread/work.
I have login sessions 'and' javascript.
I'm going to attempt the cookie method.
Wish me luck :)

danzimal:
I have compiled in the new csrf module and I can see that csrf cookies are generated and passed.
However, I never seem to get a receive_header_name match in the request headers. The following conditional always returns NULL.

mod_csrf.c line: 367
    if (NULL != (ds = (data_string *)array_get_element(con->request.headers,
                                        p->conf.receive_header_name->ptr))) {

Is the patch function supposed to append the csrf.receive-header-name?
I never get a du->key match.

mod_csrf.c line: 337
            } else if (buffer_is_equal_string(du->key, CONST_STR_LEN("csrf.receive-header-name"))) {
                PATCH(receive_header_name);

BTW, The semicolons in the original csrf.conf file caused parse errors.

jin-eld:
Phu... it's been a while...

> if (NULL != (ds = (data_string *)array_get_element(con->request.headers,
> p->conf.receive_header_name->ptr))) {

Here I'm looking at the request headers and I'm trying to get the header value by the name specified in the config, i.e. looking for conf.receive_header_name, the default is X-Csrf-Token, so here I'm actually retrieving the token that was sent back to us from the client.

> } else if (buffer_is_equal_string(du->key, CONST_STR_LEN("csrf.receive-header-name"))) {
> PATCH(receive_header_name);


Given that this was my one and only lighttpd module I do not remember the lighttpd API that well anymore, but I think this function transfers the data set in the .conf file to the plugin_config struct.

> BTW, The semicolons in the original csrf.conf file caused parse errors.

Oops... 

Our "live" production version is here btw:
https://git.digitalstrom.org/dss-oe/dss-oe/blob/master/yocto/dS/meta-digitalstrom-devel/recipes-extended/lighttpd/lighttpd/mod_csrf.patch
https://git.digitalstrom.org/dss-oe/dss-oe/blob/master/yocto/dS/meta-digitalstrom-devel/recipes-extended/lighttpd/lighttpd/fix-12840.patch

And the configuration looks like this:
https://git.digitalstrom.org/dss-oe/dss-oe/blob/master/yocto/dS/meta-digitalstrom-devel/recipes-extended/lighttpd/lighttpd/lighttpd.conf#L234

Hope that helps...

danzimal:
I reviewed your latest changes, but I still don't see where the receive_header_name is getting injected into the request header (:
I must be missing a bigger picture here. If I send a post via Postman with a header containing X-Csrf-Token, then it'll finally shows up in con->request.headers. Do you remember inserting X-Csrf-Token somewhere such as mod_fpauth?
thanks again.

jin-eld:
Ehm... I am not sure if I understand you correctly... the "receive header name" which is defined as X-Csrf-Token by default, is supposed to be injected in the client code, that's how this whole csrf thing works. Lighttpd sends out a csrf cookie to the client and expects the hash back in the request header from the client, "receive_header_name" variable itself just defines the name of that header and can be set in the config.

In my comment above, in https://github.com/lighttpd/lighttpd1.4/pull/13#issuecomment-189652273 I posted some examples of how you would globally inject those headers on the client side in jQuery and ExtJS 3... or did I misunderstand your question?

danzimal:
Ah. #13 (comment)
I got it.
Thanks again!!!

danzimal:
Apologies for keeping this thread alive with my challenges.
I am very close to successful integration.
I have ported the plugin to lighttpd 1.4.26 and have implemented our current ajax using ajaxPrefilter beforesend. That is the good news.

I am however having difficulty 'injecting' the X-csrf-token into the header from within my HTML pages.
I can use similar ajaxPrefilters and XML from within my pages and the X-csrf-token header injection always gets injected and is found and parsed by the plugin's mod_csrf_uri_handler as shown in con->request.headers.

BUT, as soon as I enable the url via csrf.protection, **the receive_header_name is no longer available to the mod_csrf_uri_handler**. If I manually set the request header in Postman, it succeeds all day long.

_(please excuse the LSP specifics and csrf was renamed)._

Some of my client examples are;
$.ajaxSetup({
  beforeSend: function(xhr) {
    xhr.setRequestHeader("X-RID", $.cookie("RID"));
  }
});
// Sends custom header
$.ajax({ url: "<?=label.url_page?>" });

and

var req = new XMLHttpRequest();
req.open("<?=request_method?>", "<?=label.url_page?>", true);
req.setRequestHeader("X-RID",  $.cookie("RID"));
req.send();

**Any thoughts and recommends are very much appreciated.**

jin-eld:
Uhm... why would you need to protect static html pages? I suggest you study https://en.wikipedia.org/wiki/Cross-site_request_forgery in detail, as I understand it you only need to protect stuff that can execute actions on your server (for instance cgi scripts).

danzimal:
Indeed I am talking about FastCGI scripts servicing GET/POST requests. I don't see a way to select POST only requests using the URL selectors for csrf.protection so I assumed I would protect overall page access.
That said, is this not reasonable? Why can't the plugin read the csrf header when protected?

jin-eld:
My configuration looks like this:
$HTTP["url"] =~ "^/(json|dss|getLatestLogs|cgi-bin(?!/logview\.hs|/downloadbackupwrapper\.sh|/restorebackup\.hs|/pushfile\.hs)|dsa(?!/html(.+)))/(.+)" {
    csrf.protection = "enable"
}

Basically I am catching all requests to applications that process json and also to the cgi-bin directory, while plain html pages are excluded, default lighttpd selectors are used.

Aren't your POST requests going to some specific URL that you can identify? Usually you don't have static html and cgi mixed...

> That said, is this not reasonable? 

What exactly? I am not sure what you are referring to...

> Why can't the plugin read the csrf header when protected?

Which one? Of a POST request from a static page? I don't know, I never tried, my use case was posting from JS via AJAX only....

Anyway, I doubt I can help you beyond of what I implemented and as I understand it does work as designed. If you want to extend the functionality further you'd have to dig into the lighttpd module API, check the mailing lists, I'm really not an expert on that topic.

gstrauss:
FYI: lighttpd config supports $HTTP["request-method"] since lighttpd 1.4.19
https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_Configuration

However, as you are well aware, any method, including GET, might be processed by a dynamic handler.  (adding this for others who might read through this post)

BTW, @danzimal, did you really backport this to 1.4.26, which was released almost 7 years ago?
Care to share your alternate token generator (for when openssl is not available)?

danzimal:
We have single LSP files implementing both GET and POST based on the request method.
gstauss, thanks for the selector link, I will certainly investigate.
I did port to 1.4.26. It wasn't too bad. The buffer.h/c helper functions had changed and the plugin skeleton was a little different. I'm wondering that maybe my older version might be contributing to my misfortune.

I currently stubbed out the token generator. This is TBD until I solidify the infrastructure :)
I will certain post my solution (assuming I don't pull in openssl crypto.

nagarajven:
Is there a mod_csrf patch for the latest lighttpd version 1.4.45 ?
Is there any reason why this did not make it to the mainline ?

jin-eld:
> Is there a mod_csrf patch for the latest lighttpd version 1.4.45 ?

@nagarajven not from my side, I was planning to update our lighttpd at some point, which would require porting the module, but did not have time for it yet.

And I guess the second question goes to the lighttpd devs...

nagarajven:
@jin-eld Thanks for your response. Please find attached below configuration relevant to csrf
#### csrf module                                                               
## 1 hour lifetime                                                             
csrf.send-cookie-name = "csrf"                                                 
#csrf.receive-header-name = "X-Csrf-Token"                                     
csrf.cookie-domain = ""                                                        
csrf.cookie-lifetime = 3600                                                    
                                                                               
$HTTP["url"] =~ "^/(json|dss|getLatestLogs|cgi-bin(?!/logview\.hs|/downloadback
    csrf.protection = "enable"                                                 
} 

When I try to visit protected url under cgi-bin my browser is receiving 403 error code and below is the log from error_log. Am I missing anything in the configuration ? 

2006-01-01 22:05:18: (mod_csrf.c.442) denying request with missing token

nagarajven:
@jin-eld Is there a way to force the browser to go to https://<ip-address>/index.html upon receiving 403 error code from web server, which will happen when device reboots while client is browsing protected url ?

nagarajven:
@jin-eld Is there a way to force the browser to go to https://172.25.50.12/index.html upon receiving 403 error code from web server, which will happen when device reboots while client is browsing protected url ?

nagarajven:
@jin-eld Thanks for your response. Please find attached below configuration relevant to csrf

# csrf module                                                                                                                                                    
# 1 hour lifetime                                                                                                                                                
csrf.send-cookie-name = "csrf"                                                                                                                                   
#csrf.receive-header-name = "X-Csrf-Token"                                                                                                                       
csrf.cookie-domain = ""                                                                                                                                          
csrf.cookie-lifetime = 3600                                                                                                                                      
                                                                                                                                                                 
$HTTP["url"] =~ "^/(json|dss|getLatestLogs|cgi-bin(?!/logview\.hs|/downloadbackupwrapper\.sh|/restorebackup\.hs|/pushfile\.hs)|dsa(?!/html(.+)))/(.+)" {         
    csrf.protection = "enable"                                                                                                                                   
}

When I try to visit protected url under cgi-bin my browser is receiving 403 error code and below is the log from error_log. Am I missing anything in the configuration ?

2006-01-01 22:05:18: (mod_csrf.c.442) denying request with missing token

nagarajven:
@jin-eld Thanks for your response. Please find attached below configuration relevant to csrf

#csrf module                                                                                                                                                     
#1 hour lifetime                                                                                                                                                 
csrf.send-cookie-name = "csrf"                                                                                                                                   
#csrf.receive-header-name = "X-Csrf-Token"                                                                                                                       
csrf.cookie-domain = ""                                                                                                                                          
csrf.cookie-lifetime = 3600                                                                                                                                      
                                                                                                                                                                 
$HTTP["url"] =~ "^/(json|dss|getLatestLogs|cgi-bin(?!/logview\.hs|/downloadbackupwrapper\.sh|/restorebackup\.hs|/pushfile\.hs)|dsa(?!/html(.+)))/(.+)" {         
    csrf.protection = "enable"                                                                                                                                   
}

When I try to visit protected url under cgi-bin my browser is receiving 403 error code and below is the log from error_log. Am I missing anything in the configuration ?

2006-01-01 22:05:18: (mod_csrf.c.442) denying request with missing token

jin-eld:
> Is there a way to force the browser to go to https://172.25.50.12/index.html upon receiving 403 error
> code from web server, which will happen when device reboots while client is browsing protected url ?

@nagarajven That depends on your web code, so I guess there is :)
The idea is not to protect static html pages, but only to protect calls that perform actions which need to be protected. So if, for instance, you are using AJAX requests to execute those actions, you can react on a 403 response in any way you like, so you can redirect the browser as you wish.

Of course you could also patch the module and return some 3xx code.

> When I try to visit protected url under cgi-bin my browser is receiving 403 error code and below is the log > from error_log. Am I missing anything in the configuration ?

Did you visit an unprotected page first? You need to acquire the cookie through an unprotected page which would usually be index.html or some page that represents the entry to your site. Alternatively, you can expect that the first request always fails, but it will acquire the new cookie for all that follow. In our scenario this happens more or less automatically, we periodically poll non critical status information which is allowed to fail, so usually if the device is rebooted in the background while the browser UI was open, this poller may fail, but it will at some point get the new cookie from the device. It's also OK for the user, because this poller also detects if the device is down or unreachable and informs the user.

I guess it depends on your use case, there are many options on how those things can be handled.

stbuehler:
Why this wasn't merged: see comment https://github.com/lighttpd/lighttpd1.4/pull/13#issuecomment-198731141 and following.

If someone wants to see such functionality upstream I suggest starting with the branch https://git.lighttpd.net/lighttpd/lighttpd1.4.git/log/?h=personal/stbuehler/mod-csrf, rebasing it and testing it, and then (if the tests were succesful) ask for merging it.

gstrauss:
Revisiting this pull request from long ago.

@jin-eld thank you for the effort you made in this pull request, and for the subsequent discussion.

Modern browsers now widely support cookies with `SameSite=Strict` so that should definitely be used with the cookie-to-header pattern.

I have completely reimplemented mod_csrf for my current development branch of lighttpd.
https://github.com/gstrauss/lighttpd1.4/tree/mod_csrf
This new implementation contains multiple options (plenty of rope) for different CSRF mitigation patterns, some of which may (or may not) be useful for integration with client-side and server-side frameworks.  Documentation is at the top of mod_csrf.c in my branch.  Though this new code has not been tested, the proof-of-concept is there and feedback on the ideas is appreciated.

jin-eld:
@gstrauss : cool, thank you for the ping! I will most likely get back to this topic within the next few months, I have not been maintaining my branch vs new releases anymore, because I switched companies ;) However I plan to use lighttpd again on another product and the CSRF protection topic will surely pop up again. Do you have any plans or timeline when the feature from your branch may go into a release?

I can't yet say when I will be able to get to it, but I will for sure check out your branch instead of rebasing my old stuff that has never been accepted, it would be good to see this feature as a part of a future lighttpd release :>

gstrauss:
(This is a general response, not aimed at you @jin-eld)

> Do you have any plans or timeline when the feature from your branch may go into a release?

Typical catch-22.  It will not be eligible for an official lighttpd release until after it has been tested by someone besides myself, and who has reasonable experience with CSRF.  ("I compiled it and it didn't crash" is not sufficient testing. :))

That said, patching a module into the source tree and then building a package for your favorite distro is not difficult, but does require a modicum of effort.

Besides, once part of an official lighttpd release, various distros move at different snail-like speeds, so getting the current lighttpd release into a distro can take anywhere from a few weeks to more than two years, and is not under the control of lighttpd developers (much to our consternation).

PR #14, opened 2015-12-27
mod_ssi: enhance support for ssi vars

fbrosson:
Try ssi_vars if ssi_cgi_env does not have a matching var name.
Allow var names to also include digits after the initial letter or underscore.


stbuehler:
Thanks! Merged in d8e028e06900b781bebfe2f52cc7254444a31716

PR #15, opened 2016-01-03
fix links to online docs in template config files

fbrosson:
Almost all template configuration files in doc/config/conf.d/ have broken links, as shown below.
This edit fixes each template config file by substituting the broken link with the correct link.
```Shell
curl --head http://www.lighttpd.net/documentation/access.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/authentification.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/cgi.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/cml.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/compress.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/dirlisting.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/evhost.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/expire.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/fastcgi.html
HTTP/1.1 404 Not Found

curl --head http://trac.lighttpd.net/trac/wiki/Docs%3AModMagnet
HTTP/1.1 301 Moved Permanently
Location: http://redmine.lighttpd.net/wiki/lighttpd/Docs:ModMagnet

curl --head http://redmine.lighttpd.net/wiki/lighttpd/Docs:ModMagnet
HTTP/1.1 301 Moved Permanently
Location: http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs:ModMagnet

curl --head http://www.lighttpd.net/documentation/Foo.html
HTTP/1.1 404 Not Found

http://www.lighttpd.net/documentation/mysqlvhost.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/proxy.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/rrdtool.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/scgi.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/secdownload.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/simple-vhost.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/status.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/trigger_b4_dl.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/userdir.html
HTTP/1.1 404 Not Found

curl --head http://www.lighttpd.net/documentation/webdav.html
HTTP/1.1 404 Not Found
```

### DIFF
--- a/doc/config/conf.d/access_log.conf
+++ b/doc/config/conf.d/access_log.conf
@@ -2,7 +2,7 @@
 ##
 ## Corresponding documentation:
 ##
-## http://www.lighttpd.net/documentation/access.html
+## http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModAccesslog
 ##
 server.modules += ( "mod_accesslog" )
 
--- a/doc/config/conf.d/auth.conf
+++ b/doc/config/conf.d/auth.conf
@@ -3,7 +3,7 @@
 ##  Authentication Module
 ## -----------------------
 ##
-## See http://www.lighttpd.net/documentation/authentification.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/docs_modauth
 ## for more info.
 ##
 
--- a/doc/config/conf.d/cgi.conf
+++ b/doc/config/conf.d/cgi.conf
@@ -3,7 +3,7 @@
 ##  CGI modules
 ## --------------- 
 ##
-## http://www.lighttpd.net/documentation/cgi.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/docs_modcgi
 ##
 server.modules += ( "mod_cgi" )
 
--- a/doc/config/conf.d/cml.conf
+++ b/doc/config/conf.d/cml.conf
@@ -3,7 +3,7 @@
 ##  CML Module 
 ## --------------- 
 ##
-## see http://www.lighttpd.net/documentation/cml.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModCML
 ##
 server.modules += ( "mod_cml" )
 index-file.names += ( "index.cml" )
--- a/doc/config/conf.d/compress.conf
+++ b/doc/config/conf.d/compress.conf
@@ -3,7 +3,7 @@
 ##  Output Compression 
 ## --------------------
 ##
-## see http://www.lighttpd.net/documentation/compress.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/docs_modcompress
 ##
 server.modules += ( "mod_compress" )
 
--- a/doc/config/conf.d/debug.conf
+++ b/doc/config/conf.d/debug.conf
@@ -3,6 +3,8 @@
 ##  Debug options
 ## ---------------
 ##
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/DebugVariables
+##
 ## Enable those options for debugging the behavior
 ##
 ## The settings can be set per location/vhost.
--- a/doc/config/conf.d/dirlisting.conf
+++ b/doc/config/conf.d/dirlisting.conf
@@ -3,7 +3,7 @@
 ##  Dirlisting Module 
 ## ------------------- 
 ##
-## See http://www.lighttpd.net/documentation/dirlisting.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModDirlisting
 ##
 
 ##
--- a/doc/config/conf.d/evhost.conf
+++ b/doc/config/conf.d/evhost.conf
@@ -3,7 +3,7 @@
 ##  Enhanced virtual hosting 
 ## -------------------------- 
 ##
-## See http://www.lighttpd.net/documentation/evhost.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModEVhost
 ##
 server.modules += ( "mod_evhost" )
 
--- a/doc/config/conf.d/expire.conf
+++ b/doc/config/conf.d/expire.conf
@@ -3,7 +3,7 @@
 ##  Expire Module 
 ## --------------- 
 ##
-## See http://www.lighttpd.net/documentation/expire.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/docs_modexpire
 ##
 server.modules += ( "mod_expire" )
 
--- a/doc/config/conf.d/fastcgi.conf
+++ b/doc/config/conf.d/fastcgi.conf
@@ -3,7 +3,7 @@
 ##  FastCGI Module 
 ## --------------- 
 ##
-## http://www.lighttpd.net/documentation/fastcgi.html
+## See http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModFastCGI
 ##
 server.modules += ( "mod_fastcgi" )
 
--- a/doc/config/conf.d/geoip.conf
+++ b/doc/config/conf.d/geoip.conf
@@ -3,6 +3,8 @@
 ##  GeoIP Module 
 ## --------------- 
 ##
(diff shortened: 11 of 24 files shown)


stbuehler:
Thanks!

Merged in 36a266ec29ce943e3485b11524161f323534e50b

fbrosson:
You are welcome, @stbuehler !
BTW, if you would like to add the "opt-out for ETag & Last-Modified headers"  feature, please do not hesitate to ask and I'll be happy to create a new PR for it. BTW, the reason why I "recycled" this PR was because I was unsure whether it would be desired or not (and I still don't know). So just drop me a line and I'll submit it. BTW, I've got another patch for mod_ssi with  `SSI_ECHO_AUTHORITY`, `SSI_ECHO_SCHEME` and, most interesting, `SSI_ECHO_SCRIPT_URI`, which makes `SCRIPT_URI` available in pages handled by the SSI module. SCRIPT_URI is really handy for adding links to the W3C Validator:
`<a href="http://validator.w3.org/check?uri=<!--#echo var="SCRIPT_URI" -->">...</a>`
This is much better than having `/check?uri=referer` because some users configure their browser to only send referers to the same site, (or completely disable referers). Having `SCRIPT_URI` makes it possible to create links to the W3C Validator that work for everyone. Moreover, `SCRIPT_URI` already exists in Apache. Just let me know if you are interested and I'll be glad to submit it in a PR. Thanks!

stbuehler:
@fbrosson Didn't realize you recycled the PR... please don't do it again :)

The additional SSI variables sound fine (if they are also supported by apache), it would be nice if you could submit them.

I'm not sure what the "opt-out for ETag & Last-Modified headers" feature would be doing (I can't see the diff anymore); but I guess it sounds fine to not send them in a dynamically created page by default.

fbrosson:
Thanks for your message and sorry about the recycled PR. I promisse I won't do that anymore.
I know that Apache defines **`SCRIPT_URI`** but I need to check whether it also defines **`SCHEME`** and **`AUTHORITY`**.
Having **`AUTHORITY`** is probably not very important because with HTTP/1.1 the server also has the information in HTTP_HOST.
Having **`SCHEME`** is handy in situations where you serve both http and https and want to display one image for http and another for https. Imagine you have images of the QR codes encoding the address of your main page in each of the protocols, http and https. Putting the QR code in the page makes it very easy to use the scan feature of modern smartphones to copy/paste an address from one screen to a mobile device, all that with the very same protocol used on the source screen. You could name your images `qrcode-http.png` and `qrcode-https.png` and then include something like **`"/>`** and mod_ssi would do the job. That's both easier and faster than inserting tests with SSI expressions to compare the contents of the SSI variable HTTPS to see if it equals "ON", if the configuration includes such a definition.
The other small advantage of having **`SCHEME`** is that you can easily "create" links that _look like_  protocol-relative links without having to rely on the browser's ability to handle true protocol-relative links (i.e. **`://`**). But I do agree that such browsers are not common and that it is not smart to use this method to create such fake protocol-relative links _just_ for these browsers.
Well, let me know which variables you would like to have and I'll create a PR with a patch for adding them. What do you think of adding both **`SCRIPT_URI`** and **`SCHEME`** ?

Regarding the opt-out feature, here's a small description of why I think it would be nice to have it.
Most of my pages look static, but I use mod_ssi to have their timestamps automaticaly displayed (with the famous and obvious  **`<!--#echo var="LAST_MODIFIED" -->`**). So for all these pages the normal behaviour of mod_ssi is perfect. But I also have some dynamic pages (e.g. a GeoIP page that displays information related to the IP address of the client), also handled by mod_ssi, for which I don't need to send any ETag or Last-Modified headers. It does not harm to send them, but since they are not necessary for such dynamic pages, it is nice to be able to opt-out on a context-based config. The feature is really simple and I'm pretty sure it can't break anything.
If you are also interested in that feature, let me know what is your favorite syntax to invert the default behaviour. Would you prefer to have this:
* ssi.omit-etag = "enable"
* ssi.omit-last-modified = "enable"

(which was my first idea), or ths:
* ssi.send-etag = "disable"
* ssi.send-last-modified = "disable"

(which was my second idea, probably slightly better) ?
Of course, I'll take care of making the change in a way that it is neutral to configurations that do not specify any setting.
In case you are interested in both of my proposals, then I guess it is best to have 2 separate PRs.
Thanks!

fbrosson:
Regarding the ETag and Last-Modified headers,  you wrote this in your last message:
> I guess it sounds fine to not send them in a dynamically created page by default

I also think it's a good idea. (It did not catch my attention immediatly, sorry).
We can use "disable" as default values for **`ssi.send-etag`** and **`ssi.send-last-modified`** and allow the user to opt-in on any context with:
* ssi.send-etag = "enable"
* ssi.send-last-modified = "enable"

Do you think the suggested keywords are fine? Would you prefer to have something else?
I'll create a PR for that with appropriate comments in the template ssi.conf file to explain this tiny change and emphasize that it only affects dynamic pages handled by mod_ssi.

Regarding the **`SCRIPT_URI`** (and maybe **`SCHEME`**), I have started making a list of which important and interesting SSI vars exist (or are missing) in Apache vs Lighttpd. I'll come back once it is finished so that you can chose which ones you would like to have added. Cheers!

stbuehler:
The `ssi.send-*` names sound fine to me. Regarding the variables I have no problem synchronizing our list with apache, as long as the variables can be easily and correctly implemented on our side.

fbrosson:
OK, cool. I'll jump into the opt-in feature very soon. Regarding the variables, **`SCHEME`** does not exist in Apache but maybe if Lighttpd adds it then Apache will follow, who knows? **`SCHEME`** is really easy to add. It can be accessible to SSI pages as SSI vars, and it can even be made available very easily as an environment variable. For **`SCRIPT_URI`** it is also easy to make it available as an SSI var (and I already have the patch for that). But it is not trivial (for me) to also make it available as an environment variable. I know Apache makes **`SCRIPT_URI`** available as an SSI var, but I'm not yet 100% sure that it also makes it available as an environment variable. I need to check that. Anyway, making it available in Lighttpd as an SSI var is a good start. And I'm pretty sure that from my patch it will be easy for several other developers to suggest a few small changes to make it also available as an environment variable. It's just a matter of calling the right routines to concatenate buffers. I knew how to do this on 1.4.35 but there were some changes since and I did not refresh my patch to make it also create the environment variable for **`SCRIPT_URI`**.

stbuehler:
If apache doesn't have a different name for `SCHEME` I guess it should be fine for now. (CGI environment probably only sets `HTTPS=on` for `https` and otherwise not at all.)

fbrosson:
After diving into some Apache dumps of dynamic pages launching ENV, some searching on the net, and finally grabbing the source tarball of Apache 2.4.18 (the current latest) and examining its source code I can confirm that it does not define any var as either http or https. As you pointed out, the only related var is **`HTTPS`** (but as you also point out it is not done the way we could do it in Lightty). And since **`SCHEME`** is not already defined in Apache and seems to be a good candidate for a var that could be set to either http or https, I think it is safe to use that name. I haven't checked yet with Nginx but I"ll try to do so.
BTW, if you grep apr_table_setn recursively you will find **`apr_table_setn(r->subprocess_env, "HTTPS", "on");`** in `modules/arch/netware/mod_nw_ssl.c`, **`apr_table_setn(e, "REQUEST_URI", original_uri(r));`** in `server/util_script.c` and several other env vars (but nothing equivalent to the **`SCHEME`** var we could add to Lightty). I'll create a PR with for these 2 vars asap.

PR #16, opened 2016-02-10
mod_cgi - use MAP_PRIVATE to mmap temporary file

gstrauss:
instead of MAP_SHARED.  flash filesystem JFFS2 does not support mmap PROT_READ MAP_SHARED, though it does support mmap PROT_READ MAP_PRIVATE
Reference: https://redmine.lighttpd.net/issues/2715

Although MAP_SHARED is preferred, CGI input body is fully collected prior to handler invoking the CGI, so the temporary file is never modified after being mapped.  Since the request input body is specific to request and is temporary file, mmap PROT_READ MAP_PRIVATE works fine.

### DIFF
--- a/src/mod_cgi.c
+++ b/src/mod_cgi.c
@@ -10,13 +10,13 @@
 #include "plugin.h"
 
 #include <sys/types.h>
+#include "sys-mmap.h"
 
 #ifdef __WIN32
 # include <winsock2.h>
 #else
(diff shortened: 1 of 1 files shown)


stbuehler:
It seems like `MAP_PRIVATE` is the better choice here, yes.

Thanks!

Merged in 3fd80ff8ec0196274eec4e0ef42718eee627cae5

PR #17, opened 2016-02-11
do not propagate sighup if 0 == server.max-workers

gstrauss:
reduce impact of sighup on child processes, such as piped loggers, by not forwarding SIGHUP signal unless server.max-workers configured

For those configuring server.max-workers, it is recommended that piped loggers be used to avoid log corruption, and then admins can avoid sending lighttpd SIGHUP as there is currently no benefit to doing so with the standard modules (beyond that of log rotation of non-piped access and error logs).

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1119,7 +1119,7 @@ int main (int argc, char **argv) {
 							 * 
 							 * we also send it ourself
 							 */
-							if (!forwarded_sig_hup) {
+							if (!forwarded_sig_hup && 0 != srv->srvconf.max_worker) {
 								forwarded_sig_hup = 1;
 								kill(0, SIGHUP);
 							}


stbuehler:
I have the feeling that someone might rely on the current behavior... perhaps some people want the pipe loggers to receive the `SIGHUP` to cycle the logfiles.

gstrauss:
SIGHUP will kill the piped logger if the logger does not catch the signal.  This is the current case for cronolog, rotatelogs (Apache), and multilog (daemontools), so I argue that the desire for lighttpd to propagate SIGHUP is unlikely to be the common case.  Instead, as I mentioned above, it is much more likely that current lighttpd configurations do send SIGHUP to lighttpd if they employ piped loggers.

Typical behavior of piped loggers is to exit if they read EOF.  Then, the server can respawn them, if desired.  Doing so would probably be more robust than blindly propagating SIGHUP, though it is not the current behavior (lighttpd does not restart piped loggers).

Also, please note that the piped loggers that I have worked with (including cronolog, rotatelogs, multilog) have their own, independent, log rotation mechanisms based on time and/or log file size, not  based on SIGHUP signal generated by an external process.

Therefore, while it might be true that "someone might rely on the current behavior", I think the patch gets rid of the unwanted "surprise" behavior of blindly propagating SIGHUP to children (killing both piped loggers and any running CGI, among other things), though this patch only does so for the 0 == server.max-workers case.

gstrauss:
To summarize: blindly propagating SIGHUP to all children processes in the process group is bad behavior.  It is not graceful to kill currently running CGI programs and other subprocesses when the goal is log rotation of mod_accesslog and error log files (and useful only if *not* using piped-loggers)

stbuehler:
You are probably right.

stbuehler:
Thanks!

Merged in 5cc061bfab42f3b81b12263b1e33355be1844fe7

PR #19, opened 2016-02-12
[mod_fastcgi] 404 for X-Sendfile file not found

gstrauss:
send 404 for X-Sendfile file not found error

(slightly modified from patch by Olaf-van-der-Spek)

reference:
  Option to map send-file file-not-found error to normal 404
  https://redmine.lighttpd.net/issues/2474


gstrauss:
I slightly modified Olaf's patch to retain the error trace when stat_cache_get_entry() fails. If it should be removed for X-Sendfile, then it should also be removed for X-Sendfile2

stbuehler:
Thanks! Applied in c80ae9b212806eea0adbc675c716ddc4f1b01777

PR #20, opened 2016-02-13
check for support for RFC 3493 IPV6_V6ONLY

gstrauss:
Reference:
  network.c does not compile on Solaris
  https://redmine.lighttpd.net/issues/2277

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -286,6 +286,7 @@ static int network_server_init(server *srv, buffer *host_token, specific_config 
 			struct addrinfo hints, *res;
 			int r;
 
+		      #ifdef IPV6_V6ONLY
 			if (s->set_v6only) {
 				val = 1;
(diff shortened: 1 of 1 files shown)


stbuehler:
`IPV6_V6ONLY` has been defined in 2003; supporting IPv6 without `IPV6_V6ONLY` seems not necessary to me.

If due to some missing header or bug the macro is not defined but actually should have been, this fix suddenly becomes a security bug, listening to an address it wasn't supposed to.

gstrauss:
Would you prefer something along the lines of:
`
#ifdef HAVE_IPV6
#ifndef IPV6_V6ONLY
#error missing IPV6_V6ONLY definition
#endif
#endif
`

gstrauss:
eh.  just let the compile fail if IPV6_V6ONLY definition is missing.  Modern OS supporting IPv6 should define IPV6_V6ONLY

PR #21, opened 2016-02-13
[mod_cgi] issue trace and exit if execve() fails

gstrauss:
(replace SEGFAULT if execve() fails)

Reference:
  sloppy error handling in mod_cgi to affect binaries
  https://redmine.lighttpd.net/issues/2302


stbuehler:
Thanks!

Merged int f23a24a263531501374f07de4b18344886f00a8c

PR #22, opened 2016-02-13
eval prev conditions in if-else before later cond

gstrauss:
The first condition which evaluates true in any if-else... condition
chain short-circuits the chain, and any remaining conditions in the
chain are marked false.

Previous conditions in if-else condition chaining must be evaluatable
(to true or false) -- must not remain in unset (not yet evaluatable)
state -- prior to evaluating later conditions.  Since any true
condition short-circuits remaining conditions, all prev conditions
must be false prior to evaluating later conditions.

Reference:
  Semantics of else clause looks strange
  https://redmine.lighttpd.net/issues/2598

### DIFF
--- a/src/configfile-glue.c
+++ b/src/configfile-glue.c
@@ -264,7 +264,8 @@ static cond_result_t config_check_cond_nocache(server *srv, connection *con, dat
 		}
 
 		/* make sure prev is checked first */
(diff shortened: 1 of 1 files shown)


stbuehler:
Thanks!

Modified, added test and applied in 1c01a42aa3bf1ff997ae1e58e53b88053f84faa8

PR #23, opened 2016-02-18
buffer_caseless_compare() edge case, signed return

gstrauss:
fixed buffer_caseless_compare() to return proper signed value
(promote char to int via (unsigned char *) cast for proper handling)

document and handle some unexpected requirements for
buffer_caseless_compare(), including operating on binary data
(buffers containing strings not ending in '\0'), as well as
on NULL args combined with size == 1 (implying '\0' ending)

reference:
  heap-use-after-free in lighttpd
  https://redmine.lighttpd.net/issues/2717

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -457,8 +457,17 @@ int buffer_caseless_compare(const char *a, size_t a_len, const char *b, size_t b
 	size_t const len = (a_len < b_len) ? a_len : b_len;
 	size_t i;
 
+	/* XXX: some existing usage of this routine expects sane behavior even
+	 * when NULL arg passed with size 1 (pretending '\0' byte exists (""))*/
+	if (a == NULL || b == NULL) {
+		return a_len < b_len ? -1 : (a_len > b_len);
+	}
+
+	/* XXX: some existing usage of this routine expects sane behavior even
+	 * when args passed do not end '\0', i.e. must compare len, not len-1 */
 	for (i = 0; i < len; ++i) {
-		unsigned char ca = a[i], cb = b[i];
+		int ca = ((unsigned char *)a)[i];
+		int cb = ((unsigned char *)b)[i];
 		if (ca == cb) continue;
 
 		/* always lowercase for transitive results */


stbuehler:
- `buffer_caseless_compare` must always be called with the "real" string length (not including terminating zeroes); the referenced issue actually was a use-after-free (i.e. corrupted memory).
- Assigning `unsigned char` from `char` values should be perfectly fine; subtracting two `unsigned char`s always ends up as `int`; the operands get promoted to `int` before the operation

I.e. I see nothing wrong so far with the current implementation.

gstrauss:
I see.  I had misread the usage of buffer_caseless_compare().  Thanks for the explanation.

As to integer promotion, yes, that's true on common platforms, but I believe is implementation-defined behavior.  A compiler could choose to store the result in an unsigned char and then do an implicit cast to int.  A simple implementation (different from what is done in lighttpd buffer_caseless_compare()) which stores the result in an unsigned char, and then promotes to int results in the wrong value.
```
    #include <stdio.h>
    int main (void)
    {
        unsigned char a = 120;
        unsigned char b = 240;
        unsigned char c = a - b;
        fprintf(stderr, "%d (a - b)\n", a - b);
        fprintf(stderr, "%d (c)\n", (int)c);
        return 0;
    }
```
results in
-120 (a - b)
136 (c)

While the current implementation in lighttpd does work, using int instead of unsigned char is likely to be slightly faster on common 32-bit and 64-bit processors because using int avoids the multiple register fills needed to promote from unsigned char to a full register for operations on unsigned char.

stbuehler:
It is not faster, as using int is exactly what the compiler will be doing. As `sizeof(unsigned char) < sizeof(int)` and the (signed) `int` can represent all `unsigned char` values, `a - b` is actually read as `((int)a) - ((int)b)`, even if you assign the result to an `unsigned char`.

This is called "integer promotion"; there is also "usual arithmetic conversions" (which didn't have any effect here because both operands got promoted to the same type): see https://www.securecoding.cert.org/confluence/display/c/INT02-C.+Understand+integer+conversion+rules

I didn't see any evidence that these rules are optional or implementation-defined behavior, so the only reason to change the code would be to make it more obvious what is happening.

stbuehler:
Do you think this would increase readability?

```patch
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -466,7 +466,8 @@ int buffer_caseless_compare(const char *a, size_t a_len, const char *b, size_t b
 		if (cb >= 'A' && cb <= 'Z') cb |= 32;
 
 		if (ca == cb) continue;
-		return ca - cb;
+		/* explicit integer promotion */
+		return ((int)ca) - ((int)cb);
 	}
 	if (a_len == b_len) return 0;
 	return a_len < b_len ? -1 : 1;
```

gstrauss:
Yes, I find the intent of
```
return (int)ca - (int)cb;
```
(with or without additional parens) more readable, but that's just my opinion.

I think your suggestion would be a good addition to the code for clarity, but it is not required.  I left this pull request closed since you fixed the root cause of https://redmine.lighttpd.net/issues/2717 in 431559e5df7dfa6edfcea0371c09339e04b402c8  Thanks!

gstrauss:
FYI, see also the commit you made to the 1.5.x branch back in 2012, where you use 'int':
<pre>
commit 24fe3d303630cf26f992f65437290d4399eeb556
Author: Stefan Bühler <stbuehler@web.de>
Date:   Sun Apr 8 08:02:51 2012 +0000

    buffer_caseless_compare: always convert letters to lowercase to get transitive results, fixing array lookups (fixes #2405)
    
    git-svn-id: svn://svn.lighttpd.net/lighttpd/trunk@2829 152afb58-edef-0310-8abb-c4023f1b3aa9
</pre>

stbuehler:
Just that it missed the step going through `unsigned char` and I have no idea anymore why I did it.
Anyway, fixed in 2abdf0143be8510c0d21ca498cba9684d052802c

PR #25, opened 2016-02-22
check get_http_*_name() for NULL return

gstrauss:
reference:
  https://redmine.lighttpd.net/issues/2583


stbuehler:
Applied in 733ce382030bdf4fe362b24da1ab7728aca8d11f

PR #26, opened 2016-02-22
validate return value from strtol, strtoul

gstrauss:
reference:
  https://redmine.lighttpd.net/issues/2564


stbuehler:
Thanks! Applied first commit in f5453290b7742443da98728cacfeeca42aafa7f6 with some small changes:
- ignore trailing whitespace in configparser, but otherwise fail hard (no warning) on conversion failure
- use `buffer_string_length()` instead of `->used`
- add some `{` ... `}`

PR #27, opened 2016-02-23
warn if server.upload-dirs has non-existent dirs

gstrauss:
Warn at startup if any dirs in server.upload-dirs do not exist.
Take server.chroot into account, if set.

reference:
  https://redmine.lighttpd.net/issues/2508

### DIFF
--- a/src/configfile.c
+++ b/src/configfile.c
@@ -1290,17 +1290,18 @@ int config_set_defaults(server *srv) {
 
 	if (srv->srvconf.upload_tempdirs->used) {
 		buffer * const b = srv->tmp_buf;
-		size_t used;
-		b->used = 0;
+		size_t len;
 		if (!buffer_string_is_empty(srv->srvconf.changeroot)) {
 			buffer_copy_buffer(b, srv->srvconf.changeroot);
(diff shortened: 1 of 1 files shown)


stbuehler:
I'd like to get rid of any direct uses of `->used` in a buffer; please use the functions from `buffer.h` instead (`buffer_string_is_empty`, `buffer_string_length`, `buffer_string_set_length`)

gstrauss:
updated patch to use buffer_string_length() and buffer_string_set_length()

(FYI, the patch still tests a->used to see if array is empty.  I did not see an equivalent array_is_empty() interface.)

stbuehler:
Thanks! Squashed and applied in b4a4afdaf7f873a3aa08f98e248f93542dc72161

I needed the wrapper functions for buffer, as `->used` sometimes included a trailing null character and sometimes didn't; also both "0" and "1" mark empty strings, but only "0" is an empty buffer (a distinction needed in some places). The wrapper functions make sure that if `->ptr` is not NULL the buffer is always null terminated.

This extra logic is not needed for arrays - here `->used` is always exactly the number of entries.

PR #28, opened 2016-02-23
configparser.y: warn when dup keys in array merge

gstrauss:
prior behavior for merging arrays of arrays silently discarded
second array. New behavior issues warning.

prior behavior for merging arrays of integers silently discarded
second integer value. New behavior issues warning.

prior behavior for merging arrays of strings would join strings
with ", ", which might be appropriate for HTTP header list fields,
but not for config parsing.  New behavior issues warning.

This addresses the issue reported in:
  "weird outcome for duplicate alias map"
  https://redmine.lighttpd.net/issues/2685

### DIFF
--- a/src/configparser.y
+++ b/src/configparser.y
@@ -91,7 +91,12 @@ data_unset *configparser_merge_data(data_unset *op1, const data_unset *op2) {
       for (i = 0; i < src->used; i ++) {
         du = (data_unset *)src->data[i];
         if (du) {
-          array_insert_unique(dst, du->copy(du));
+          if (du->is_index_key || buffer_is_empty(du->key) || !array_get_element(dst, du->key->ptr)) {
+            array_insert_unique(dst, du->copy(du));
+          } else {
+            fprintf(stderr, "WARNING: duplicated key '%s'; "
+                            "ignoring subsequent values\n", du->key->ptr);
+          }
         }
       }
       break;


glensc:
:+1: at least for warning, but you seem to break existing behavior. not sure what it is, but you yourself mentioned header merging

gstrauss:
I can not imagine a situation where joining an array of strings with ", " results in a valid, desirable config.  In https://redmine.lighttpd.net/issues/2685 it produces a broken config.  The behavior change here produces a valid config of the first item, and a warning for duplicated keys.

Similarly, it is very likely a config error -- and a warning should be issued -- in the cases where the second item was previously being silently discarded.  The resulting config in this case is the same as before, with the addition of the warning.

glensc:
yes, i know about 2685, i reported it

gstrauss:
FYI: it appears this issue was addressed in the 1.5.x branch back in 2005.

<pre>
commit ed43eed7f29312065f786af684c5c00df2607a3c
Author: Jan Kneschke <jan@kneschke.de>
Date:   Sat Apr 9 16:33:19 2005 +0000

    duplication config-vars and duplicate array key are a error now

    git-svn-id: svn://svn.lighttpd.net/lighttpd/trunk@267 152afb58-edef-0310-8ab
b-c4023f1b3aa9
</pre>

gstrauss:
@stbuehler committed change in https://redmine.lighttpd.net/projects/lighttpd/repository/revisions/3103, modifying it to throw an error instead of only issuing a warning.  That works!

(FYI: my original reasoning for making the patch with a warning was to have zero impact on existing configs with this problem, besides the extra trace.)

PR #29, opened 2016-02-24
[proxy] accept LF delimited headers, not just CRLF

gstrauss:
reference:
  "Reverse proxy return 0 byte for some pages"
  https://redmine.lighttpd.net/issues/2594


stbuehler:
Thanks! Applied in 0aa2ea74e3fcc181fd70e2968320a9803e685bcc

PR #33, opened 2016-02-27
[mod_compress] case-insensitive content-codings

gstrauss:
case-insensitive parse of Accept-Encoding for content-codings

reference
  lighttpd doesn't support case-insensitive for content-coding values ??
  https://redmine.lighttpd.net/issues/2645


stbuehler:
Thanks!

Applied in 75e4859a1ba0211e99951bae933f643b03eb012f

PR #34, opened 2016-02-27
[mod_webdav] allow Depth: Infinity lock on file

gstrauss:
(still not supporting Depth: Infinity on directories)

reference
  "Saving file to webdav mapped drive fails"
  https://redmine.lighttpd.net/issues/2296

patch by mstorsjo submitted as part of feature request
  https://redmine.lighttpd.net/issues/1953


gstrauss:
commited in ab829cee

PR #36, opened 2016-03-01
server.error-handler new directive for error pages

gstrauss:
server.error-handler preserves HTTP status error code when error page is static, and allows dynamic handlers to change HTTP status code when error page is provided by dynamic handler.  server.error-handler intercepts all HTTP status codes >= 400 except when the content is generated by a dynamic handler (cgi, ssi, fastcgi, scgi, proxy, lua).  The request method is unconditionally changed to GET for the request to service the error handler, and the original request method is later restored (for logging purposes).  request body from the original request, if present, is discarded.
    
server.error-handler is somewhat similar to server.error-handler-404, but server.error-handler-404 is now deprecated, intercepts only 404 and 403 HTTP status codes, and returns 200 OK for static error pages, a source of confusion for some admins.  On the other hand, the new server.error-handler, when set, will intercept all HTTP status error codes >= 400.  server.error-handler takes precedence over server.error-handler-404 when both are set.
    
NOTE: a major difference between server.error-handler and the now-deprecated server.error-handler-404 is that the values of the non-standard CGI environment variables REQUEST_URI and REDIRECT_URI have been swapped.  Since REDIRECT_STATUS is the original HTTP status code, REDIRECT_URI is now the original request, and REQUEST_URI is the current request (e.g. the URI/URL to the error handler).  The prior behavior -- which reversed REQUEST_URI and REDIRECT_URI values from those described above -- is preserved for server.error-handler-404.
    
Additionally, REDIRECT_STATUS is now available to mod_magnet, which continues to have access to request.uri and request.orig_uri.
    
See further discussion at https://redmine.lighttpd.net/issues/2702 and https://redmine.lighttpd.net/issues/1828

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -81,6 +81,7 @@ NEWS
   * [core] open fd when appending file to cq (fixes #2655)
   * [config] server.listen-backlog option (fixes #1825, #2116)
   * [core] retry tempdirs on partial write, ENOSPC (fixes #2588)
+  * [core] compile with upcoming openssl 1.1.0 release (fixes #2727)
 
 - 1.4.39 - 2016-01-02
   * [core] fix memset_s call (fixes #2698)
--- a/doc/config/conf.d/mime.conf
+++ b/doc/config/conf.d/mime.conf
@@ -8,8 +8,19 @@
 ## https://redmine.lighttpd.net/projects/lighttpd/wiki/Mimetype_assignDetails
 
 ##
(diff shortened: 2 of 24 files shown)


gstrauss:
mod_magnet and mod_rewrite might also modify con->request.uri, so I'll add a patch that has them set REDIRECT_URI, too.

PR #37, opened 2016-03-02
update .gitignore, add .gitattributes

gstrauss:
more user-friendly settings for development in git


stbuehler:
I'd rather not use this. Build in a subdirectory, add it to your local excludes and you should be fine.

The attributes look completely unrelated to the project; I don't see why we would want them locally, and why they are not getting added to some system default if they are that important.

gstrauss:
Regarding .gitignore: I develop in git.  This is committed to my development tree.  (At a previous employer of mine, using .gitignore and developing in-tree is how 100+ developers worked, too, so this is more than simply my personal preference.)  Others who develop in git will probably find this useful in reducing noise, as opposed to using 'git status -uno'.  What potential problems do you see occurring due to these additions to .gitignore?  (In short: what's the harm?)

Regarding .gitattributes, it is not the default in git due to historical reasons, but having .gitattributes is considered best practices when using git.  Again,what would be the harm in adding it?  It enforces sanity with line endings in the git repository, whether people are using unix-style or Windows-style line endings.
https://help.github.com/articles/dealing-with-line-endings/
http://stackoverflow.com/tags/gitattributes/hot

stbuehler:
I develop in git too (obviously). I use various build systems, and build in subdirectories (both cmake and autotools support this. scons has a fixed subdirectory). This is why only those files which always are generated in the source directory by autotools are on the ignore list.

To account for local stuff I therefore have this in in `.git/info/exclude`:

    *.patch
    *build/
    CMakeLists.txt.user

Ignoring every file which might be created by a in-tree build is a high maintenance burden, and I don't see any upside. I don't care how many people do in-tree building, doesn't make them right :)
(I'd be open to adding `*build` and `CMakeLists.txt.user` because I don't see a maintenance burden here).

Similar arguments go for the attributes. I don't think this recommended list is going to stay the same forever, so you probably would want to update it from time to time, but I simply don't see why we would need all of them in the first place.

Best practices... I don't see them here: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/

gstrauss:
Regarding .gitattributes:
> Best practices... I don't see them here: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/

Written by a developer at Github, http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/ is an excellent article about the history of managing line endings in git and the recommendation to use .gitattributes.

Regarding .gitignore:
> Ignoring every file which might be created by a in-tree build is a high maintenance burden,

I disagree with both those statements.  How many pull requests have you gotten to modify .gitignore?  I do not know how you can conclude that it is a high maintenance burden for this repository, though I do understand the sentiment that you are expressing, which is that you do not want to be making frequently updates to .gitignore.  I agree with the sentiment.

> and I don't see any upside.

For an example of an upside, how about "more inclusiveness"?  Besides the unsubstantiated projection of maintenance burden, what is the harm for ignoring obvious build artifacts?  I think we agree that these build artifacts should not be checked into the git repository.

> I don't care how many people do in-tree building, doesn't make them right :)

Having managed source control and written multiple build systems for a company with 3000+ developers, I can tell you one thing with absolute certainty: there is more than one "right" (read: "functional") way to do development.  Some might be more optimal than others, and many have trade-offs that make them better in some ways and worse in others, but I think that it is shortsighted to assume that "only one way is right", and that by implication everyone else is "doing it wrong". 

While I can use .git/info/exclude in each copy of the repository that I have (on multiple machines), when I am working with others on a project, having .gitignore and .gitattributes as part of the repository is one good way to help with consistency.

I'm going to close this pull request, since you've expressed disinterest in these optional features, but I do hope that you'll reconsider.

gstrauss:
Some additional links to .gitattributes articles:
http://makingsense.github.io/migration-to-git/3-working-with-git/gitattributes-file.html
http://makingsense.github.io/migration-to-git/3-working-with-git/dealing-with-line-endings.html

PR #38, opened 2016-03-03
411 Length Required if Transfer-Encoding: chunked

gstrauss:
send 411 Length Required if Transfer-Encoding: chunked request body
(lighttpd does not currently support request body transfer-codings)

x-ref:
  "PROPFIND status code not propogated in proxy"
  https://redmine.lighttpd.net/issues/631

This allows lighttpd frontend to proxy to backend Apache using mod_dav_svn to serve svn repositories.

### DIFF
--- a/src/request.c
+++ b/src/request.c
@@ -1131,7 +1131,15 @@ int http_request_parse(server *srv, connection *con) {
 		}
 		break;
 	default:
-		/* the may have a content-length */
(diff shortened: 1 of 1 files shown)


stbuehler:
Thanks! Applied in 06d3c754403b2df99775641e05999ea9b2081618

PR #40, opened 2016-03-04
[config] detect standard modules in modules list

gstrauss:
Fix detection of standard modules in modules list.
This patch fixes spurious warnings such as:
2016-03-03 21:52:32: (plugin.c.186) Cannot load plugin mod_indexfile more than once, please fix your config (lighttpd may not accept such configs in future releases)
when "mod_indexfile" is listed in server.modules = ( ... )

### DIFF
--- a/src/configfile.c
+++ b/src/configfile.c
@@ -1174,6 +1174,7 @@ int config_read(server *srv, const char *fn) {
 	}
 
 	if (NULL != (modules = (data_array *)array_get_element(srv->config, "server.modules"))) {
+		size_t i;
 		data_string *ds;
 		data_array *prepends;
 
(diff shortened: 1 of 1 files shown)


gstrauss:
FYI: found similar commit to 1.5.x branch from 2007:
<pre>
commit 649c5e41a3f15ade3b1ff9af43d97dff455354a1
Author: Jan Kneschke <jan@kneschke.de>
Date:   Tue May 8 12:28:01 2007 +0000

    fixed the loading for default modules if they are specified explicitly

    git-svn-id: svn://svn.lighttpd.net/lighttpd/trunk@1836 152afb58-edef-0310-8a
bb-c4023f1b3aa9
</pre>

stbuehler:
I backported the patch from trunk instead in 9ae781368532bd831aeac1a769c8ca27a60e9736

PR #41, opened 2016-03-05
consistent inclusion of config.h at top of files

gstrauss:
x-ref:
  "wrong position for #include <sys/types.h> in network_backends.h"
  https://redmine.lighttpd.net/issues/2073

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -1,3 +1,5 @@
+#include "first.h"
+
 #include "array.h"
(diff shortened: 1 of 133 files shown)


stbuehler:
Thanks. Applied (minus the md5 redirects) in 8abd06a7ff2a1505f9c9a79196074a6e805efa1b

PR #42, opened 2016-03-07
more careful parse of $SERVER["socket"] config str

gstrauss:
detect unix domain socket path earlier
detect IPv6 addr without port (might contain ':' within addr, e.g. [::])

This commit fixes some things so that diff is smaller for discussion in
  "Enable IPv6 conf"
  https://redmine.lighttpd.net/issues/2204

(BTW, this diff is easier to review when ignoring leading whitespace changes.)

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -161,7 +161,6 @@ static int network_server_init(server *srv, buffer *host_token, specific_config 
 	int val;
 	socklen_t addr_len;
 	server_socket *srv_socket;
-	char *sp;
 	unsigned int port = 0;
 	const char *host;
 	buffer *b;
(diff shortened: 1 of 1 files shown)


src/network.c
```diff
@@ -194,36 +193,44 @@ static int network_server_init(server *srv, buffer *host_token, specific_config
...
+		/* host is a unix-domain-socket */
+		is_unix_domain_socket = 1;
+	} else {
+		/* ipv4:port
+		 * [ipv6]:port
+		 */
+		size_t len = buffer_string_length(b);
+		char *sp = NULL;
+		if ((b->ptr[0] == '[' && b->ptr[len-1] == ']') || NULL == (sp = strrchr(b->ptr, ':'))) {
+			if (NULL == sp) {
```
stbuehler: I think this condition (`NULL == sp`) is always true; if `(b->ptr[0] == '[' && b->ptr[len-1] == ']')` the second part of `||` doesn't get evaluated and `sp` is still `NULL`; otherwise the second part was true and `NULL == (sp = ...)`.

src/network.c
```diff
@@ -194,36 +193,44 @@ static int network_server_init(server *srv, buffer *host_token, specific_config
...
+			*(sp-1) = '\0';
+			host++;
 
-		s->use_ipv6 = 1;
-	}
+			s->use_ipv6 = 1;
+		}
 
-	*(sp++) = '\0';
+		if (*sp == ':') {
```
stbuehler: I don't see how this could be false.

gstrauss:
@stbuehler: please look at the additional patch at the bottom of https://redmine.lighttpd.net/issues/2204 and you'll see why some of the extraneous code in this patch exists.  It is preparing for the follow-up patch.

stbuehler:
It is fine to motivate patches with future changes, but I think the current code should make sense too. The patch in issue 2204 doesn't require the first "special case", and could handle the second part nicer. I will modify the patches and apply them.

stbuehler:
Applied in d85bdab43ff115763deb6f5cb292f650397fe5f4

PR #43, opened 2016-03-20
command line opt to shutdown after idle time limit

gstrauss:
-i <secs>  graceful shutdown after <secs> of inactivity

Option might be used with applications such as git instaweb.  While git instaweb does have command line options of its own to [start,stop,restart], some may find it convenient to configure git instaweb to start lighttpd with a default idle time limit, after which lighttpd will gracefully shut itself down without any further action from the user.

x-ref:
  "[PATCH] support -i <secs> idle timeout option"
  https://redmine.lighttpd.net/issues/2696
  original request and patch submitted by mackyle.  thx.

### DIFF
--- a/doc/lighttpd.8
+++ b/doc/lighttpd.8
@@ -39,6 +39,9 @@ Test the configuration file for syntax errors, load and initialize modules, and 
 \fB\-D\fP
 Do not daemonize (go into background). The default is to daemonize.
 .TP 8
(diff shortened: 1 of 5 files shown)


gstrauss:
[experimenting with github; attempting to reopen]

gstrauss:
[not working; will re-file pull request]

gstrauss:
[ok; must restore branch in github interface before reopening pull request]

PR #44, opened 2016-03-20
lighttpd -tt performs preflight startup checks

gstrauss:
lighttpd -t loads config file and performs syntax check
lighttpd -tt (new) performs preflight startup checks,
  including loading and initializing modules, but skipping any
  potentially destructive actions which might affect an already
  running server (separate instance).  These currently include:
  - skipping pidfile modification
  - skipping bind() to network sockets
  - skipping open of error and access logs

x-ref:
  "typo in config file did not show up in a config-test"
  https://redmine.lighttpd.net/issues/411

### DIFF
--- a/src/base.h
+++ b/src/base.h
@@ -510,6 +510,7 @@ typedef struct {
 	buffer *breakagelog_file;
 
 	unsigned short dont_daemonize;
+	unsigned short preflight_check;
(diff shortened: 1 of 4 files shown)


stbuehler:
Thanks! Applied in 292309f88b3ebf083ea69202943add6e907d3321

PR #45, opened 2016-03-20
mimetype.xattr-name global config option

gstrauss:
For backwards compatibility with existing lighttpd configs, default is
  mimetype.xattr-name = "Content-Type"

Those who wish to use the freedesktop.org definition of xattr mimetype
can set the following in the global lighttpd config:
  mimetype.xattr-name = "user.mime_type"

x-ref:
  "Migrate to freedesktop.org definition of xattr mimetype"
  https://redmine.lighttpd.net/issues/2631

### DIFF
--- a/src/base.h
+++ b/src/base.h
@@ -541,6 +541,7 @@ typedef struct {
 	} stat_cache_engine;
 	unsigned short enable_cores;
 	unsigned short reject_expect_100_with_417;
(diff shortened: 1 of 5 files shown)


stbuehler:
Thanks! Applied in d17d48e01e384fe85eb2a148f28e81539a02913e.

PR #46, opened 2016-03-23
Update base64.c

mnaza:
If `in_tupre_reminder` is `0`, then `out_tuple_reminder` is `0` too, and `4*full_tuples` can be equal to `4*full_tuples + out_tuple_remainder`.


stbuehler:
Thanks! Applied in 7b983ae05467c06e6cf06d7ba6dad76490a13206

PR #47, opened 2016-03-27
[mod_ssi] config ssi.conditional-requests

fbrosson:
Summary:
A new SSI directive, «**`ssi.conditional-requests`**», allows to inform lighttpd which SSI pages should be considered as cacheable and which should not. In particular, the **`ETag`** & **`Last-Modified`** headers will only be sent for those SSI pages for which the directive is enabled.

Long description:
**`ETag`** and **`Last-Modified`** headers were being sent for all SSI pages, regardless of whether they were cacheable or not. And yet, there was no cache validation at all for any SSI page.
This commit fixes these two minor issues by adding a new directive, «**`ssi.conditional-requests`**», which allows to specify which SSI pages are cacheable and which are not, and by adding cache validation to those SSI pages which are cacheable. And since sending ETags for non-cacheable documents is not appropriate, they are no longuer computed nor sent for those SSI pages which are not cacheable.
Regarding the **`Last-Modified`** header for non-cacheable documents, the standards allow to either send the current date and time for that header or to simply skip it. The approach chosen is to not send it for non-cacheable SSI pages. **`ETag`** and **`Last-Modified`** headers are therefore only sent for an SSI page if «**`ssi.conditional-requests`**» is enabled for that page.

The «**`ssi.conditional-requests`**» directive can be enabled or disabled globally and/or in any context. It is disabled by default.

An index.shtml which only includes determinist SSI commands such as:
**`<!--#echo var="LAST_MODIFIED"-->`**
is a trivial example of a dynamic SSI page that is cacheable.

Refs: [RFC7232](https://tools.ietf.org/html/rfc7232) & [RFC7234](https://tools.ietf.org/html/rfc7234).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -74,6 +74,7 @@ NEWS
   * [mod_dirlisting] class for dir <tr> (fixes #2304)
   * [core] define __STDC_WANT_LIB_EXT1__ (fixes #2722)
   * [core] setrlimit max-fds <= rlim_max for non-root (fixes #2723)
+  * [mod_ssi] config ssi.conditional-requests
 
 - 1.4.39 - 2016-01-02
   * [core] fix memset_s call (fixes #2698)
--- a/doc/config/conf.d/ssi.conf
+++ b/doc/config/conf.d/ssi.conf
@@ -1,6 +1,6 @@
 #######################################################################
 ##
-##  Server Side Includes 
+##  Server Side Includes
 ## -----------------------
 ##
 ## See http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ModSSI
@@ -11,6 +11,35 @@ server.modules += ( "mod_ssi" )
 ## which extensions should be ran through mod_ssi.
 ##
 ssi.extension              = ( ".shtml" )
+
+##
+## The  ssi.conditional-requests  directive only affects requests
+## handled by the SSI module and allows to declare which SSI pages
+## are cacheable and which are not. This directive can be enabled
+## or disabled globally and/or in any context.
+##
+## As the name of this directive suggests, conditional requests will
+## be handled appropriately for any SSI page for which the directive
+## is enabled. In particular, the "ETag" and "Last-Modified" headers
+## will both be sent. Conversely, these headers will NOT be sent for
+## pages for which the directive is disabled.
+##
+## The directive should be set to "enable" ONLY for requests that are
+## known to generate cacheable documents. An SSI page which only
+## includes contents from other static files and/or which uses SSI
+## commands that produce predictable output (e.g. the echo command
+## for the LAST_MODIFIED variable) is likely to be cacheable.
+##
+## The directive should be set to "disable" for ALL other documents,
+## that is, for SSI pages which depend on non-predictable input such
+## as (but not limited to) output from ssi exec commands, data from
+## the client's request headers (other than the request URI), or any
+## other non constant input such as the current date or time, the
+## client's user-agent, etc...
+##
+## Disabled by default.
+##
+#ssi.conditional-requests = "enable"
 
 ##
 #######################################################################
--- a/src/mod_ssi.c
+++ b/src/mod_ssi.c
@@ -107,6 +107,7 @@ SETDEFAULTS_FUNC(mod_ssi_set_defaults) {
 	config_values_t cv[] = {
 		{ "ssi.extension",              NULL, T_CONFIG_ARRAY, T_CONFIG_SCOPE_CONNECTION },       /* 0 */
 		{ "ssi.content-type",           NULL, T_CONFIG_STRING, T_CONFIG_SCOPE_CONNECTION },      /* 1 */
+		{ "ssi.conditional-requests",   NULL, T_CONFIG_BOOLEAN, T_CONFIG_SCOPE_CONNECTION },     /* 2 */
 		{ NULL,                         NULL, T_CONFIG_UNSET, T_CONFIG_SCOPE_UNSET }
 	};
 
@@ -121,9 +122,11 @@ SETDEFAULTS_FUNC(mod_ssi_set_defaults) {
 		s = calloc(1, sizeof(plugin_config));
 		s->ssi_extension  = array_init();
 		s->content_type = buffer_init();
+		s->conditional_requests = 0;
 
 		cv[0].destination = s->ssi_extension;
 		cv[1].destination = s->content_type;
+		cv[2].destination = &(s->conditional_requests);
 
 		p->config_storage[i] = s;
 
@@ -1078,7 +1081,7 @@ static int mod_ssi_handle_request(server *srv, connection *con, plugin_data *p) 
 		response_header_overwrite(srv, con, CONST_STR_LEN("Content-Type"), CONST_BUF_LEN(p->conf.content_type));
 	}
 
-	{
+	if (p->conf.conditional_requests) {
 		/* Generate "ETag" & "Last-Modified" headers */
 		time_t lm_time = 0;
 		buffer *mtime = NULL;
@@ -1093,6 +1096,13 @@ static int mod_ssi_handle_request(server *srv, connection *con, plugin_data *p) 
 
 		mtime = strftime_cache_get(srv, lm_time);
 		response_header_overwrite(srv, con, CONST_STR_LEN("Last-Modified"), CONST_BUF_LEN(mtime));
+
+		if (HANDLER_FINISHED == http_response_handle_cachable(srv, con, mtime)) {
+			/* ok, the client already has our content,
+			 * no need to send it again */
+
+			chunkqueue_reset(con->write_queue);
+		}
 	}
 
 	/* Reset the modified time of included files */
@@ -1112,6 +1122,7 @@ static int mod_ssi_patch_connection(server *srv, connection *con, plugin_data *p
 
 	PATCH(ssi_extension);
 	PATCH(content_type);
+	PATCH(conditional_requests);
 
 	/* skip the first, the global context */
 	for (i = 1; i < srv->config_context->used; i++) {
@@ -1129,6 +1140,8 @@ static int mod_ssi_patch_connection(server *srv, connection *con, plugin_data *p
 				PATCH(ssi_extension);
 			} else if (buffer_is_equal_string(du->key, CONST_STR_LEN("ssi.content-type"))) {
 				PATCH(content_type);
+			} else if (buffer_is_equal_string(du->key, CONST_STR_LEN("ssi.conditional-requests"))) {
+				PATCH(conditional_requests);
 			}
 		}
 	}
--- a/src/mod_ssi.h
+++ b/src/mod_ssi.h
@@ -17,6 +17,7 @@
 typedef struct {
 	array *ssi_extension;
 	buffer *content_type;
+	unsigned short conditional_requests;
 } plugin_config;
 
 typedef struct {


fbrosson:
Hi @stbuehler (and @gstrauss )!
In case you are planning to release 1.4.40 soon I thought you might wish to merge this really simple commit. It won't hurt nor break anything ;-)
Thanks in advance for reviewing and, maybe, merging too!

gstrauss:
Why two separate directives instead of a single directive (e.g. ssi.conditional-headers = "enable")

Reference for ETag and Last-Modified: https://tools.ietf.org/html/rfc7232
RFC 7232 uses "SHOULD" to recommend sending both ETag and Last-Modified, when possible.

More generally, what is the goal of this patch?  (What problem is it solving?)  Why not send these headers with server-side includes?  Yes, there is some minor overhead to calculating them, but do you have an additional reason for not sending the headers?

The ETag and Last-Modified header by themselves do not enable caching.  Instead, they enable a client to make a conditional request using the information provided in those headers obtained in a previous request.

mod_ssi does not currently call http_response_handle_cachable(), so even if a client provides these headers in a request, the headers are not used by the current implementation of mod_ssi.  mod_ssi will always send new content (and, regardless, it still needs to generate the content to calculate the ETag and Last-Modified).  Now, if mod_ssi is enhanced to conditionally send 304 Not Modified based on ETag or Last-Modified, then I can see also adding a directive to enable or disable handling the conditional request for ssi, e.g. with a new directive like ssi.conditional-requests = "enable".

fbrosson:
Thanks for your message (and for the pointer).
> Why two separate directives instead of a single directive (e.g. ssi.conditional-headers = "enable")?

Well, that's because each of these headers can be sent or not sent, independently of whether the other one is also being sent or not. Merging the 2 directives would reduce flexibility. Moreover, although this is not recommended, it should be possible to send only the Last-Modified header and not bother computing the ETag.

The RFC states that the server «_SHOULD send a Last-Modified value if it is feasible to send one._»
My understanding is that sending that header for a content that depends on non constant input (such as the current date&time, the client's IP address or user-agent, or the system's statistics) would not be appropriate because setting the Last-Modified header to the date&time the request was processed brings no value and can even confuse the client as it can think the page was not dynamically generated. But I have to admit that I'm not sure whether or not there is concensus on what I'm saying here.

The RFC also states that the server «_SHOULD send an entity-tag validator unless it is not feasible to generate one._»
My reasoning here is simitar to the one for the Last-Modified header: It does not make sense to compute the ETag for something that is not constant.

The RFC concludes with: «_In other words, the preferred behavior for an origin server is to send both a strong entity-tag and a Last-Modified value in successful responses to a retrieval request._»

My interpretation here is that since no client should expect to be sent these headers for all requests it is not forbidden to not send them or to send only one of them. I would say that a server software should not add limitations based on what is _only_ recommended. The responsability for not following a recommendation should be on the webadmin, not on the server software. No one will ever blame the software for being flexible as long as its default behaviour fully complies with the standards.
The default configuration of lighttpd already follows the recommendations and will continue to do so even if the patch is merged. This is because the default settings for ssi.send-etag and ssi.send-last-modified are both "enable".

> More generally, what is the goal of this patch?

I have a private server (for family access) which has several index.shtml pages for which I send the ETag and  Last-Modified headers. But a few other index.shtml pages have content which changes. That's the case for a "print-user-agent" or "print-env" page, as well as for a geoip page which replies with info taken from the free GeoIP databases for the client's IP. Sending the headers for those pages is not necessary, and we might even say that it is not desirable at all.
So what I do is that I set both directives to enable (well, with the current patch it is already the default value) at the top level and then, for selected pages I override the global directive with "disable".
I agree that the CPU time saved by not sending the headers is neglegible, but I think it makes sense to allow a sort of opt-out on a page by page basis, i.e. for selected configuration contexts.

I hope I have convinced you ;-) Thanks again for your feedback.

gstrauss:
Summary:
* sorry, I am not convinced about value of this patch
* might be able to make an argument that mod_ssi should never send ETag and Last-Modified unless explicitly enabled with a config directive

RFC 7232 is for conditional requests.  ETag and Last-Modified are headers which contribute to conditional requests.  RFC 7234 is for caching.  These are not the same thing, and I think you have misunderstood the difference.

Sorry, I am not convinced this patch has any value beyond besides the marginal cost of generating the headers, but does add a different marginal cost of config merging for every request which touches mod_ssi, whether or not it is an SSI request.  Since the marginal costs are all marginal, my conclusion is that this patch is unnecessary.

mod_ssi currently follows the RFC 7232 recommendations (not requirement) to send ETag and Last-Modified if mod_ssi can reasonably do so.  While it is true that for some SSI content, the ETag is insufficient to uniquely identify the generated content, as I explained above, mod_ssi currently does not process these headers if provided in conditional requests, so there is little impact to sending them with every response.

Sometimes having too many configuration knobs can be as bad as having too few.

If mod_ssi is enhanced to process conditional requests, there should be a way to disable it, and it should be disabled by default.  If it is disabled by default, then I could be convinced that mod_ssi should not generate ETag or Last-Modified when conditional processing is disabled for URLs handled by mod_ssi.  I still maintain that such a config switch should be a single directive.


On the other hand, with all that having been said above, I might be convinced if you wish to argue that mod_ssi should never generate ETag and Last-Modified (unless explicitly enabled by admin via a directive) since server-side includes are dynamic (or could be dynamic), and ETag and Last-Modified as currently implemented in mod_ssi are insufficiently to reasonably be used in _all_ conditional requests handled by mod_ssi.

fbrosson:
I are right on everything. (And I was making some confusions indeed.)
> mod_ssi should never generate ETag and Last-Modified (unless explicitly enabled by admin via a directive).

I buy your suggestion ;-)
Regarding the merged directive, would **`ssi.conditional-headers`** be used only to decide whether or not the ETag and Last-Modified headers should be sent?
If yes, do you think **`ssi.send-conditional-headers`** would be a better name? (I admit it is a bit long, but it is slightly more expressive.)
I plan to update the patch in order to merge the two directives and take **`ssi.conditional-headers`** for the new name (or **`ssi.send-conditional-headers`** if everyone is OK with the longer name), change the default to "disable" and adjust ssi.conf accordingly. (I'll also make a big change in the commit message, of course).
The only _problem_ with changing the default to "disable" is that a big notice will probably need to be made to warn webadmins that they should review their config to enable the new SSI directive for their "static-like" SSI dynamic pages. BTW, this was the reason why I though, initialy, that it was better to use "enable" as the default value. But you have convinced me!

gstrauss:
I think it should be a single directive <code>ssi.conditional-requests = "enable"</code>.  When the directive is enabled, mod_ssi should generate ETag and Last-Modified _and_ should call http_response_handle_cachable() after generating the response.  If the directive is not enabled, then ETag should not be generated and Last-Modified should either be generated to current time (RFC recommendation), or Last-Modified should not be generated.

I can help you with the part of the patch which calls http_response_handle_cachable() if you like.  This is the part which gives meaning to <code>ssi.conditional-requests = "enable"</code>.

fbrosson:
I _think_ I got it right (thanks to you kind guidance), but I'm not sure. Am I calling **`http_response_handle_cachable`** the correct way?
I haven't tested (nor compiled it) yet. And I also need to read curl's man pages to figure out how to issue the appropriate conditional requests to check if everything is working as expected.
I also need to update the commit message and to improve the short text in **`doc/config/conf.d/ssi.conf`** which I just changed to:
```
## conditional-requests should only be enabled for documents for which
## both the "ETag" and "Last-Modified" headers should be sent.
## This only affects dynamic content handled by this module.
## Default is disable.
```
I think my (new) description is not appropriate because it does not mention what happens in relation to the **`http_response_handle_cachable`** call. What would write instead?
Thanks again for all your feedback ;-)

gstrauss:
I took a quick look at the patch and it looks good (I have not compiled or tested it yet)

As for text description, we want to make it clear to the reader when it is appropriate (or not) to enable.  Here's some text which you can cut-n-paste and mangle as you please:

> ssi.conditional-requests = "disable" (default)
> ssi.conditional-requests should be enabled only for documents which are cachable based only on modification times of the documents included by the SSI.  ssi.conditional-requests should not be enabled if the content changes based on anything else, such as (but not limited to) changes based ssi exec, changes based on expansion of request headers from the client, changes based on expansion of SSI directives other than SSI include, etc.

fbrosson:
Thanks @gstrauss for your text description. It helped me a lot to improve **`doc/config/conf.d/ssi.conf`**.
Well, I added quite some text around and between yours, so I don't know if what I did is really an improvement... I sometimes tend to be too verbose. But here the reason is that the default value for the new directive can have quite some consequences if non experienced users don't pay enough attention to the topic. Thanks again for your feedback!

gstrauss:
I like most of your expanded explanation.

I think you can omit (delete):
> +## An important consequence is that most clients (if not all) won't
> +## even try to cache pages for which these headers were not sent.
> +## WARNING: This could result in drastic network traffic increase
> +## comming from webcrawiers of other bots.
> +## IMPORTANT NOTICE: lighttpd releases prior to 1.4.40 bahaved as
> +## if the directive was enabled, but since enabling it by default
> +## for all SSI documents was not a sane option it was disabled by
> +## default.

"comming" is spelled "coming"
"bahaved" is spelled "behaved"

In any case, there is little impact.  RFC-compliant clients and bots would not serve cached content without revalidating it with a conditional request, and mod_ssi did not validate conditional requests (until this patch).  In fact, there may actually be (slightly) _reduced_ network traffic since bots might not send the extra headers in conditional requests if a prior response did not include ETag or Last-Modified.

fbrosson:
Thanks @gstrauss for your comments! I removed the 8 lines of text you have spotted since you have convinced me once more ;-)
BTW, I'm still laughing at my own spelling mistakes ;-)
Regarding the commit message (which is also in the description of this PR) please do not hesitate to let me know if you have any remarks or suggestions! Thanks!

gstrauss:
Thanks for being so receptive to the feedback. :)

Similar comments apply to the commit message as to your recent documentation update.  No reason to cause people alarm.  The "IMPORTANT WARNING" section and "Extreme care" paragraphs can be removed.

The default behavior of the new directive is safe.

fbrosson:
I've dropped the scary paragraphs in the commit message ;-)
> Thanks for being so receptive to the feedback. :)

It's always a pleasure to get feedback from smart and experienced contributors ;-)

fbrosson:
@gstrauss, I've made a few extra changes to the commit message as well as to the comments in **`doc/config/conf.d/ssi.conf`** and I think you'll agree with me that it is even better than before ;-)
BTW, I noticed that the return value of **`mod_ssi_handle_request()`** is only tested against 0, so I replaced the **`return HANDLER_FINISHED;`** (after the call to **`http_response_handle_cachable(..)`**) by a call to **`chunkqueue_reset(con->write_queue);`**.
@stbuehler, thanks to the feedback from @gstrauss I was able to better understand the topic, improve the patch I had in mind initially and improve the commit message and the commends in ssi.conf.

**I haven't tested the patch yet** (with the change about the cache validation) but I hope I'll find the time to do so. If you think the patch can be accepted then please let me know; this will give me more reasons to test the patch quickly ;-)
Thanks!

gstrauss:
@fbrosson: yes, this patch can be accepted.  Would you please rebase against tip of master?

Also, please correct spelling of "deterministic" (not "determinist")

fbrosson:
@gstrauss: Cool, thanks!
**IMPORTANT** I haven't tested this bloc yet, sorry, but I will do so today:
```C
		if (HANDLER_FINISHED == http_response_handle_cachable(srv, con, mtime)) {
			/* ok, the client already has our content,
			 * no need to send it again */

			chunkqueue_reset(con->write_queue);
		}
```
> Would you please rebase against tip of master?

Yes, of course. Done. The date in the commit log is «Wed Apr 13 09:07:12 2016 +0000».
> Also, please correct spelling of "deterministic" (not "determinist") and remove any trailing whitespace (e.g. the line above Default:...)

I fixed the spelling of «deterministic» in the commit message.
For the trailing whitespaces, I removed the line above the «**`## Default is disable.`**» and then I realised I had misunderstood your request. So I added the line back, but I don't know if there was a trailling whitespace in it before. I found one trailling whitespace in line 3 of ssi.conf (and got rid of it).
I also changed the «**`## Default is disable.`**» into a «**`## Disabled by default.`**» which we already have in mime.conf. I guess this is slightly better than before, isn't it?
BTW, I noticed that there are several files in **`doc/config/conf.d/`** which have trailling whitespaces. 
So if you wish I can fix them all, either in the same commit or in a separate one, at your choice. (I don't know which is better, nor if it is a good or bad idea to fix them within this PR).
Thanks for your kind feed-back and guidance ;-)

fbrosson:
Oops, there was a typo in the text I had added in **`ssi.conf`**. It's fixed now:
 «**`In particual`**» → «**`In particular`**»

fbrosson:
Oops, one more typo fix in **`ssi.conf`**: «**`Conveversely`**» → «**`Conversely`**».
I think there aren't any left.

fbrosson:
Good news: all tests are OK. I've set-up two SSI pages, one with **`ssi.conditional-requests`** disabled and the other with it enabled, and then I used **`curl`** to request each of these pages in various ways:
a) without any time condition;
b) with **`--time-cond local-file-with-an-old-time-stamp`**
c) with **`--time-cond local-file-with-a-recent-time-stamp`**

As expected:
1) the SSI page which has its  **`ssi.conditional-requests`** disabled never produces «**`ETag:`**» or «**`Last-Modified:`**» headers, and the page is always sent (unless **`--head`**» is also passed, of course).
2) the SSI page which has its  **`ssi.conditional-requests`** enabled always produces the «**`ETag:`**» and «**`Last-Modified:`**» headers and, depending on how the request is sent,
 * for cases a and b: the «**`HTTP/1.1 200 OK`**» status and the «**`Content-Length:`**» header.
 * for case c: the «**`HTTP/1.1 304 Not Modified`**» status and **no** «**`Content-Length:`**» header at all.

I did not try sending conditional request based on an ETag but I'm sure everything is also OK since the **`http_response_handle_cachable()`** is called for all kinds of conditional requests.

So I think everything is OK. Thanks in advance for the merge ;-)

gstrauss:
Thank you, and thank you particularly for testing.

May I ask that you update the commit message from 'Long descripton:' to 'Long description:' and rebase once more against tip of master?  Also, please change the commit title (the top oneline) to a shorter "[mod_ssi] config ssi.conditional-requests".  These are not critical, but are for neatness to keep the merge graph a bit clearner and so that github can reflect your commit instead of me editing it and changing the commit SHA.  I was catching up on some pending commits, so master moved quite a bit yesterday.  This is now at the top of the list!

Now that this is the next merge, please also update NEWS to match the title (the oneline)
> * [mod_ssi] config ssi.conditional-requests

at the bottom of the list of changes that will be in 1.4.40.  Thanks!

gstrauss:
Great!  Merged into my local master.  I expect it to be published this coming Monday.  Thanks!

fbrosson:
Done. BTW, I'm very honoured to be the owner of this commit!
Regarding the tests, I need to confess that I'm a little bit worried about what would happen with Keep-alives because I only though about that recently. I'll try to test if it is possible to keep the connection open after a «**`HTTP/1.1 304 Not Modified`**» on a cacheable SSI page. I'll try to do that with curl. If it's not possible then I'll use telnet. (I'll do that today.)
Hmm, I just realised that if this commit is the last to enter release 1.4.40 then I'll be the proud owner of the very first and very last changes of that release, as well as the one obout **`SCRIPT_{URI,URL}`**. So I was wondering if it would be OK to add a "(thx fbrosson)" to the right of the line I just added in NEWS (like @stbuehler did for the first line ;-) (Don't worry, I won't cry if you say no ;-)

fbrosson:
Keep-alives are working fine with all SSI pages. I just checked it with telnet, using headers grabbed from the _Live HTTP headers_ Firefox extension.
Lightty was closing my connections because I had adeed **`server.max-keep-alive-requests = 0`** in my config a long time ago to test something... and had completely forgotten to remove it.
I'm looking forward to seeing 1.4.40 released ;-) Thanks again for the kind feed-back you gave me.

gstrauss:
FYI: It is probably a month or two before we'll consider tagging and releasing 1.4.40.  There are other pending changes that need some time for review and integration.

Your changes (in this pull request) are scheduled to be published to the master branch on Monday.  The commit has your name on it, so I didn't think adding your own (thx fbrosson) was needed.  That said, I do plan to start a CONTRIBUTORS file in the near future.

Thanks for your contribution!

fbrosson:
OK, thanks. BTW, don't worry about the release date. I already build from the Git repo (for my private use).
Regarding the CONTRIBUTORS file, please don't bother with that topic if you were not already planning to do so before this PR.

I've got a tiny change request that might be interesting to add: On Apache quotes are not necessary around VAR_NAME in «**`<!--#echo var="VAR_NAME" -->`**» so it is possible to use «**`<!--#echo var=VAR_NAME -->`**» (and probably «**`<!--#echo var=VAR_NAME-->`**» without the space before the «**`-->`**», but I would need to check that).
Whereas with lightty one needs to keep the quotes around VAR_NAME otherwise the whole SSI echo command (with its tag markers) gets sent totaly unparsed. I'm sure it wouldn't be very difficult to drop that requirement. If you think it's a good idea then I can try to prepare a patch for it. What do you think?

gstrauss:
Making it easier for others to port from Apache to lighttpd (or even the other way) sounds useful.  You might also consider trying to support single quotes, as I remembered coming across https://redmine.lighttpd.net/issues/1768 "mod_ssi doesn't accept single quotes".  There are definitely people who would welcome this.

I am happy to take a look at what you put together for this or other mod_ssi enhancements.  Of course, please issue a separate pull request, as this one will be closed on Monday. :)

fbrosson:
Hmm, I had also though about single quotes but forgot to mention it. It is definitely a good idea to also support them.
I hope I'll be able to prepare a small, robust and elegant patch during the next few weeks. And, of course, I'll put it in a separate PR..

gstrauss:
(Off topic to this ticket, but to extend the syntax, you'll have to be creative with the regex in mod_ssi.c since adding more captures will change the capture numbering.  Keep in mind it might be easier to capture the results of an alternation, and later remove surrounding single- or double- quotes before using the capture string)

fbrosson:
Thanks @gstrauss for the merge as well as for your valuable feedback and kind guidance!

gstrauss:
@fbrosson: FYI: I pushed a dev branch to https://git.lighttpd.net/lighttpd/lighttpd1.4.git personal/gstrauss/master with support for no-quotes, single-quotes, or double-quotes around SSI param values.  (Caution, that branch might be force-pushed without notice)

fbrosson:
Thanks @gstrauss for the info and pointer. I'll test it and let you know if I find any problem.

fbrosson:
I think I found a bug in the mod_ssi of the current HEAD (https://github.com/lighttpd/lighttpd1.4/commit/a5fcfee6fca39ddc951dda92e6c0665981acd890). I've almost finished an index.shtml test page. What would be the most appropriate place and way to upload it?

PR #48, opened 2016-03-29
fixes for some warnings/errors reported by static analysis tools

gstrauss:


stbuehler:
- one of the reasons we have the sockaddr union is so we don't have to cast it all the time
- I actually liked the `CvtHex` wrapper function - it keeps cast to a single place
- `umask` before `mkstemp` - i don't think this is actually worth it. i think it is unlikely that someone running such old libc is updating lighty, but not the libc - every libc should be using mode 0600 now (glibc 2.06 is from 1998; POSIX.1-2008 requires it). maybe there is a way to check for POSIX.1-2008 and #error otherwise
- iterate using `char **` - I don't see the point. The new version looks worse.

gstrauss:
> one of the reasons we have the sockaddr union is so we don't have to cast it all the time

The 'plain' member of the lighttpd-specific sock_addr union is a (struct sockaddr), which has a certain size.  The size passed into getsockname() via &our_addr_len might be larger than the size of (struct sockaddr), and so static analysis tools might reasonably warn of potential overflow of storage size (struct sockaddr).  Passing a pointer to the sock_addr union points to storage at least as large as the size passed into getsockname() via &our_addr_len.

The prototype of getsockname is:
<code>int getsockname(int sockfd, struct sockaddr *addr, socklen_t *addrlen);</code>

The cast of storage to (struct sockaddr *) is common usage with BSD sockets, your opinion notwithstanding.

BTW, instead of the lighttpd-specific sock_addr union, industry standard practice is to use struct sockaddr_storage, which has been around at least as long as 2004, and might pre-date the lighttpd-specific sock_addr union.  http://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/socket.h.html

gstrauss:
> I actually liked the CvtHex wrapper function - it keeps cast to a single place

You changed the patch to add CvtHex() back, but did not add a parameter to pass length.  You also removed the change to pass buffer size in http_auth_digest_generate_nonce().  While good compilers will warn about char (*Hex)[33] argument of insufficient storage size, passing a pointer to a differently sized array is not an error in C, as it is just a pointer.

**When you modify a patch, please indicate that you have modified it with an addendum to the commit message**  The patch originated from my pull request, so I appreciate the credit.  However, since the patch committed differs (in more than trivial whitespace), I think that the commit message should note that @stbuehler modified the patch.  Thanks.

gstrauss:
> umask before mkstemp - i don't think this is actually worth it. i think it is unlikely that someone running such old libc is updating lighty, but not the libc - every libc should be using mode 0600 now (glibc 2.06 is from 1998; POSIX.1-2008 requires it). maybe there is a way to check for POSIX.1-2008 and #error otherwise

I agree and hope that it is uncommon to use such an old library.

If versioned symbols were used, it might be possible to have a binary fail to load if the versioned symbol is not available.

gstrauss:
> iterate using char ** - I don't see the point. The new version looks worse.

<code>char **environ;</code> is the declaration of the global.  I merely shortened it and iterate by pointer on stack instead of array index of a global.  Both are semantically equivalent and take the same number of lines of code, and the same number of words.  Given that likeness, what is the meaning of "The new version looks worse"?  Your statement is hardly constructive.

Since you did say "I don't see the point", I'll try to explain.  'environ' is a global.  While you might assume that an optimizing compiler might cache the value in a register, it does not have to, and might not do so for a global such as environ, which might be modified by other threads.  Each access to 'environ' for calculation of address using pointer and array index might re-access the global value.  This can be more expensive -- but never less expensive -- than accessing the global once, storing the address in a local stack variable, and using the stack variable.  (Of course, if another thread modifies the memory, there is danger of invalidating the saved value, but that is also true of access via environ[i] and is only one of many potential problems of accessing unprotected global data which might be modified by multiple threads.)

gstrauss:
When reviewing the commits in this pull request, please keep in mind that the purpose of this pull request is to reduce noise generated by static analysis tools.  That potentially means changing some code to look slightly different, but have the same results, in order for static analysis tools to have better information about what is happening.

gstrauss:
> umask before mkstemp - i don't think this is actually worth it.

Keeping the previous comment in mind, lighttpd 1.4.x is not threaded, so it is not an issue that umask() temporarily modifies global program state, and the cost of executing umask() twice is negligible compared to mkstemp() creating a temporary file.  Setting umask() would eliminate the warning from static analysis tools about unsafe umask() with mkstemp().

Yes, I still agree that people should not be using really old glibc <= 2.06.

gstrauss:
re-did the environ iteration via array-index notation instead of via pointer increment.  This should look more like the original code, with the change being to shorten spelling of 'environ' to 'e' after storing environ pointer in e.

gstrauss:
@stbuehler: it looks like your modification of cc0adf9 patch did not include the change to the signature of mod_dirlisting.c:http_list_directory_sizefmt() to pass the size of the buffer, so the call to li_itostrn() _assumes_ the buffer size.

...I haven't gone line-by-line comparing my submission to what ended up being committed in my name, but perhaps I should.

gstrauss:
I will integrate the remaining suggestions, except for umask() with mkstemp() and leaving CvtHex() unchanged.

PR #49, opened 2016-03-30
http_chunk_append_file opens fd when adding to cq

gstrauss:
and defers calculation of content length until response is finished

This reduces race conditions pertaining to stat() and then (later)
open(), when the result of the stat() was used for Content-Length
or to generate chunked headers.

Note: this does not change how lighttpd handles files that are modified
in-place by another process after having been opened by lighttpd --
don't do that.  This *does* improve handling of files that are
frequently modified via a temporary file and then atomically renamed
into place.

mod_fastcgi has been modified to use http_chunk_append_file_range() with
X-Sendfile2 and will open the target file multiple times if there are
multiple ranges.

Note: (future todo) not implemented for chunk.[ch] interfaces used by
range requests in mod_staticfile or by mod_ssi.  Those uses could lead
to too many open fds.  For mod_staticfile, limits should be put in place
for max number of ranges accepted by mod_staticfile.  For mod_ssi,
limits would need to be placed on the maximum number of includes, and
the primary SSI file split across lots of SSI directives should either
copy the pieces or perhaps chunk.h could be extended to allow for an
open fd to be shared across multiple chunks.  Doing either of these
would improve the performance of SSI since they would replace many file
opens on the pieces of the SSI file around the SSI directives.

x-ref:
  "Serving a file that is getting updated can cause an empty response or incorrect content-length error"
  https://redmine.lighttpd.net/issues/2655

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -203,6 +203,27 @@ void chunkqueue_reset(chunkqueue *cq) {
 	cq->bytes_out = 0;
 }
 
+void chunkqueue_append_file_fd(chunkqueue *cq, buffer *fn, int fd, off_t offset, off_t len) {
+	chunk *c;
+
+	if (0 == len) {
+		close(fd);
+		return;
+	}
+
+	c = chunkqueue_get_unused_chunk(cq);
+
+	c->type = FILE_CHUNK;
+
+	buffer_copy_buffer(c->file.name, fn);
+	c->file.start = offset;
+	c->file.length = len;
+	c->file.fd = fd;
+	c->offset = 0;
+
+	chunkqueue_append_chunk(cq, c);
+}
+
 void chunkqueue_append_file(chunkqueue *cq, buffer *fn, off_t offset, off_t len) {
 	chunk *c;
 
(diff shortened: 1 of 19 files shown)


src/chunk.c
```diff
@@ -203,6 +203,24 @@ void chunkqueue_reset(chunkqueue *cq) {
 	cq->bytes_out = 0;
 }
 
+void chunkqueue_append_file_fd(chunkqueue *cq, buffer *fn, int fd, off_t offset, off_t len) {
+	chunk *c;
+
+	if (0 == len) return;
```
stbuehler: either `fd` leaks here or it leaks later - define and document whether `chunkqueue_append_file_fd` is supposed to own the passed `fd` and handle it accordingly.

gstrauss:
Good catch.  The chunkqueue owns the fd passed into it.  Force-pushed updated commit.

<pre>
diff --git a/src/chunk.c b/src/chunk.c
index 3f00e34..3527152 100644
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -206,7 +206,10 @@ void chunkqueue_reset(chunkqueue *cq) {
 void chunkqueue_append_file_fd(chunkqueue *cq, buffer *fn, int fd, off_t offset, off_t len) {
        chunk *c;
 
-       if (0 == len) return;
+       if (0 == len) {
+               close(fd);
+               return;
+       }
 
        c = chunkqueue_get_unused_chunk(cq);
 
</pre>

PR #50, opened 2016-03-30
[config] server.listen-backlog option

gstrauss:
See doc/config/lighttpd.conf for explanation of listen() backlog queue

Additionally, mod_fastcgi and mod_scgi backend servers can now also be
configured with separate listen-backlog settings per server

x-ref:
  "add server.listen-backlog option instead of hard-coded value (128 * 8) for listen()"
  https://redmine.lighttpd.net/issues/2116
  "Don't disable backend when overloaded"
  https://redmine.lighttpd.net/issues/1825


gstrauss:
(I'll update the wiki doc for mod_fastcgi and mod_scgi if this patch is accepted)

PR #51, opened 2016-04-01
[fastcgi] [scgi] check for unix socket path dup

gstrauss:
error if unix socket path is duplicated

(does not check across modules, but will detect duplicated unix socket
 paths within fastcgi directives, and separately, duplicated unix socket
 paths within scgi directives)

x-ref:
  "Should warn if two FastCGI servers point to same socket file"
  https://redmine.lighttpd.net/issues/319

### DIFF
--- a/src/mod_fastcgi.c
+++ b/src/mod_fastcgi.c
@@ -1167,6 +1167,24 @@ static int fcgi_spawn_connection(server *srv,
 
 #endif /* HAVE_FORK */
 
+static int unixsocket_is_dup(plugin_data *p, size_t used, buffer *unixsocket) {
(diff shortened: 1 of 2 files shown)


stbuehler:
Duplicate socket paths are perfectly fine, just spawning on the same is usually a bad idea.

gstrauss:
If the config file specifies "bin-path" with "socket", then the server might attempt to spawn it.  Please see updated patch which adds check for !buffer_string_is_empty(host->bin_path) before checking for duplicated unix socket path.

gstrauss:
committed in 97556d99

PR #52, opened 2016-04-04
support IPv6 in $HTTP["remote-ip"] CIDR cond match

gstrauss:
x-ref:
  "Matching IPv6 addresses with $HTTP["remoteip"]"
  https://redmine.lighttpd.net/issues/2706

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -579,6 +579,16 @@ add_executable(test_base64
 )
 add_test(NAME test_base64 COMMAND test_base64)
 
+add_executable(test_configfile
+	test_configfile.c
+	buffer.c
+	array.c
+	data_string.c
+	keyvalue.c
+	log.c
+)
+add_test(NAME test_configfile COMMAND test_configfile)
+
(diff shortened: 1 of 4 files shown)


stbuehler:
I'm not a fan of matching IPv4 as v4-mapped-to-v6 addresses. As long as `server.set-v6only` is not disabled (it is enabled by default) the mapped addresses should not appear anyway.

gstrauss:
server.set-v6only = "enabled" means that the $SOCKET directive will only listen() to IPv6 traffic on that socket.  It allows a separate $SOCKET directive to listen() to the IPv4 address.  Whether or not server.set-v6only is enabled, this patch addresses how $HTTP["remote-ip"] condition will match addresses.  If my $SOCKET directives are set to listen to both INADDR_ANY and in6addr_any, then I may need to create *two* $HTTP["remote-ip"] conditions to match an IPv4 address and the equivalent IPv4-mapped IPv6 address.

gstrauss:
BTW, forced-pushed an update to the commit to add test_configfile to CMakeLists.txt.

PR #53, opened 2016-04-07
untangle overly complex control flow logic

gstrauss:
for dynamic handlers mod_cgi, mod_fastcgi, mod_scgi, mod_proxy
(mod_cgi control flow logic simplification began in a prior commit)

- connection state machine calls the subrequest handler
- subrequest handler sets up the connection to the backend
  and registers fdevent handler to handle backend events
- fdevent handler handles backend events and then schedules
  a call to connection state machine
- when retrying an alternate backend, backend state is reset
  and then response state is reset so that the connection state
  machine will call back into the subrequest handler to retry

Dynamic handlers no longer directly modify connection state
(calls to connection_set_state() from dynamic handlers were removed)

Dynamic handlers no longer reset con->physical.path, and they
preserve con->mode when retrying alternate backends.  This is done
to skip repeated processing in response.c:http_response_prepare()

While this patch increases consistency in control flow handling,
there is more work to be done that can further improve upon this.

x-ref:
  "handle-req time too long"
  https://redmine.lighttpd.net/issues/1149


gstrauss:
Added follow-on commit:

defer reading request body until handle subrequest
    
read request body right before calling subrequest handler,
allowing request to be handled prior to reading request body,
e.g. to send 401 Unauthorized response when authentication is required
    
(In the future, this might move into each dynamic handler which supports
 request body (mod_cgi, mod_fastcgi, mod_proxy, mod_scgi, mod_webdav) so
 that each dynamic handler can choose whether or not to buffer request
 body or to stream request body to backend as request body is received.)
    
keep-alive is disabled if request body has not been completely read
prior to sending response
    
Note: vast majority of lines changed in this commit are merely from
moving connection_handle_read() and connection_handle_read_ssl() from
connections.c to connections-glue.c.  The routines were moved verbatim,
without any changes, except to make connection_handle_read() public by
removing 'static'.
    
x-ref:
  "HTTP 401 Unauthorized only sent back after full POST request is read"
  https://redmine.lighttpd.net/issues/2541

gstrauss:
The reason connection_handle_read_post_state() is separate from connection_handle_read_state() is that if the handlers become responsible for triggering read of request body, then connection_handle_read_post_state(), connection_handle_read(), and connection_handle_read_ssl() can move to connections-glue.c, and connection_handle_read() would need to be made public in connections.h.

gstrauss:
force-pushed with small change to mod_scgi since handle_connection_close and connection_reset hooks should always return HANDLER_GO_ON so that all plugins have a chance to clean up.

Otherwise, resouce leaks were theoretically possible when SCGI request results in a 404 error handler, which runs a different dynamic handler.

PR #57, opened 2016-04-18
fallback to lseek()/read() if mmap() fails (#fixes 2666)

gstrauss:
fallback to lseek()/read() if mmap() fails (#fixes 2666)
e.g. when mmap() is used on lighttpd-controlled temporary files
used POST request body (mod_cgi) and PUT file upload (mod_webdav)

replace use of stream_open() on potentially untrusted files
(protect against SIGBUS if a file is modified while map is read)
Note: stream.[ch] may be removed in a future release
For now, stream.[ch] will read entire file into memory if mmap fails
and so it should only be used on trusted files, e.g. config files.

http_auth basic and digest files are typically small and so buffered
stdio fopen(), fgets(), fclose() will likely be approximately as fast
as mmap.

mod_dirlisting header and readme files are typically small and so
open(), read(), close() will typically be approximately as fast as mmap

mod_ssi will likely be much faster, now buffering SSI page construction
rather than a potentially huge number of file open() calls, one for each
tiny chunk of text between SSI directives.

mod_webdav COPY and MOVE may be slower due to removal of mmap, but are
now more resilient to partial writes.

x-ref:
  "handle filesystems without mmap() support"
  https://redmine.lighttpd.net/issues/2666
  "WebDAV upload-> mmap failed: operation not permitted"
  https://redmine.lighttpd.net/issues/962
  "Lighttpd 1.4.20 Crash (SIGBUS in mod_compress)"
  https://redmine.lighttpd.net/issues/1879
  "Crash SIGBUS"
  https://redmine.lighttpd.net/issues/2391

### DIFF
--- a/src/http_auth.c
+++ b/src/http_auth.c
@@ -67,29 +67,24 @@ static void CvtHex(const HASH Bin, char (*Hex)[33]) {
 handler_t auth_ldap_init(server *srv, mod_auth_plugin_config *s);
 
 static int http_auth_get_password(server *srv, mod_auth_plugin_data *p, buffer *username, buffer *realm, buffer *password) {
-	int ret = -1;
-
 	if (buffer_is_empty(username) || buffer_is_empty(realm)) return -1;
 
 	if (p->conf.auth_backend == AUTH_BACKEND_HTDIGEST) {
-		stream f;
-		char * f_line;
+		FILE *fp;
+		char f_user[1024];
 
 		if (buffer_string_is_empty(p->conf.auth_htdigest_userfile)) return -1;
 
-		if (0 != stream_open(&f, p->conf.auth_htdigest_userfile)) {
+		fp = fopen(p->conf.auth_htdigest_userfile->ptr, "r");
+		if (NULL == fp) {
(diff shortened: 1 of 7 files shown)


gstrauss:
@fbrosson: may I request an extra set of eyes on this pull request?  I made a large number of changes to mod_ssi and could use some help testing in another environment (yours). :)  Changes in this pull request should make it much easier to replace the SSI regex with a simple routine to parse the parameters of the SSI directive whether single-quotes, double-quotes, or no quotes are used.  The new function mod_ssi_parse_ssi_stmt() is passed the exact SSI statement string and length (not '\0' terminated), so I modified the regex to anchor it "^...$".  Still, the regex leaves much to be desired and I do not think that making it more complex is the best answer. Also, the current regex does not handle multiple consecutive backslashes correctly in some cases.

fbrosson:
Yep, of course you may! I'll have a look at the changes, build and test everything I can, and especially the SSI part. I can't promisse anything but I'll try to come back by the end of the week (or before) with any findings. BTW, thanks for your last message with the suggestions regarding the single quotes enhancement, and for the current changes which will help fixing https://redmine.lighttpd.net/issues/1768.

PR #63, opened 2016-05-11
response.c: set SSL_CLIENT_VERIFY & SSL_CLIENT_S_DN

mackyle:
SSL_CLIENT_VERIFY is set to "NONE", "SUCCESS" or "FAILED:reason".
This is compatible with Apache's mod_ssl variable of the same name.

SSL_CLIENT_S_DN is set to the oneline version of the client certificate
subject's distinguished name and may be used as a setting for the
ssl.verifyclient.username config option.  When Apache's mod_ssl is
configured to use 'FakeBasicAuth' it uses the SSL_CLIENT_S_DN value for
the username (that ultimately may end up in REMOTE_USER).  The value
that will be set for SSL_CLIENT_S_DN may be determined using the
`openssl x509 -noout -subject -in <cert.pem>` command.

Signed-off-by: Kyle J. McKay

### DIFF
--- a/src/response.c
+++ b/src/response.c
@@ -136,16 +136,51 @@ static void https_add_ssl_entries(connection *con) {
 	X509 *xs;
 	X509_NAME *xn;
 	X509_NAME_ENTRY *xe;
+	data_string *ds_cv;
+	char *s_dn;
+	long vr;
 	int i, nentries;
 
+	if (NULL == (ds_cv = (data_string *)array_get_element(con->environment, "SSL_CLIENT_VERIFY"))) {
+		if (NULL == (ds_cv = (data_string *)array_get_unused_element(con->environment, TYPE_STRING))) {
+			ds_cv = data_string_init();
+		}
+		buffer_copy_string(ds_cv->key, "SSL_CLIENT_VERIFY");
+	}
+
 	if (
-		SSL_get_verify_result(con->ssl) != X509_V_OK
+		(vr = SSL_get_verify_result(con->ssl)) != X509_V_OK
 		|| !(xs = SSL_get_peer_certificate(con->ssl))
 	) {
+		buffer_copy_string(ds_cv->value, (vr == X509_V_OK && !xs) ? "NONE" : "FAILED:bad verify result");
+		array_insert_unique(con->environment, (data_unset *)ds_cv);
 		return;
 	}
 
+	buffer_copy_string(ds_cv->value, "SUCCESS");
+	array_insert_unique(con->environment, (data_unset *)ds_cv);
 	xn = X509_get_subject_name(xs);
+	if (NULL != (s_dn = X509_NAME_oneline(xn, NULL, 0))) {
+		data_string *envds;
+		if (NULL == (envds = (data_string *)array_get_unused_element(con->environment, TYPE_STRING))) {
+			envds = data_string_init();
+		}
+		buffer_copy_string(envds->key, "SSL_CLIENT_S_DN");
+		buffer_copy_string(envds->value, s_dn);
+		OPENSSL_free(s_dn);
+		if (buffer_is_equal(con->conf.ssl_verifyclient_username, envds->key)) {
+			data_string *ds;
+			if (NULL == (ds = (data_string *)array_get_element(con->environment, "REMOTE_USER"))) {
+				if (NULL == (ds = (data_string *)array_get_unused_element(con->environment, TYPE_STRING))) {
+					ds = data_string_init();
+				}
+				buffer_copy_string(ds->key, "REMOTE_USER");
+				array_insert_unique(con->environment, (data_unset *)ds);
+			}
+			buffer_copy_buffer(ds->value, envds->value);
+		}
+		array_insert_unique(con->environment, (data_unset *)envds);
+	}
 	for (i = 0, nentries = X509_NAME_entry_count(xn); i < nentries; ++i) {
 		int xobjnid;
 		const char * xobjsn;


gstrauss:
x-ref: https://redmine.lighttpd.net/issues/2693

src/response.c
```diff
@@ -136,16 +136,51 @@ static void https_add_ssl_entries(connection *con) {
...
+			if (NULL == (ds = (data_string *)array_get_element(con->environment, "REMOTE_USER"))) {
+				if (NULL == (ds = (data_string *)array_get_unused_element(con->environment, TYPE_STRING))) {
+					ds = data_string_init();
+				}
+				buffer_copy_string(ds->key, "REMOTE_USER");
+				array_insert_unique(con->environment, (data_unset *)ds);
+			}
+			buffer_copy_buffer(ds->value, envds->value);
+		}
+		array_insert_unique(con->environment, (data_unset *)envds);
```
gstrauss: Is there any reason why this is not further up with the code which fills in envds?

src/response.c
```diff
@@ -136,16 +136,51 @@ static void https_add_ssl_entries(connection *con) {
...
+		if (buffer_is_equal(con->conf.ssl_verifyclient_username, envds->key)) {
+			data_string *ds;
+			if (NULL == (ds = (data_string *)array_get_element(con->environment, "REMOTE_USER"))) {
+				if (NULL == (ds = (data_string *)array_get_unused_element(con->environment, TYPE_STRING))) {
+					ds = data_string_init();
+				}
+				buffer_copy_string(ds->key, "REMOTE_USER");
+				array_insert_unique(con->environment, (data_unset *)ds);
+			}
+			buffer_copy_buffer(ds->value, envds->value);
```
gstrauss: What might be the implications of overwriting the existing value of REMOTE_USER (if "REMOTE_USER" already existed in con->environment prior to SSL client cert validation)?  Will SSL client cert validation occur prior to mod_auth?  If so, this might be a moot point.  I haven't looked closely yet.

gstrauss:
I added a couple comments inline to the diff.  Please review.

I presume that you are using Apache names of SSL_* vars as the convention to follow.
https://httpd.apache.org/docs/2.4/mod/mod_ssl.html
Is that the case?  I have not yet checked other popular web servers (e.g. nginx) or proxies (e.g. haproxy) or caching proxies (e.g. varnish) (and I am sure there are many other popular servers in each of those categories).

While on the subject, of variables, how do you think the values should be encoded?  Please see the paragraphs below the env var table in the Environment Variables section, as well as the SSLOptions LegacyDNSStringFormat.  What do you think?

gstrauss:
@mackyle have you had a chance to review my feedback?  Also, as noted in #62, I plan to review all the TLS-related tickets together so that solutions will not conflict.

gstrauss:
@mackyle have you had a chance to review my feedback? Also, as noted in #62, I plan to review all the TLS-related tickets together so that solutions will not conflict.

gstrauss:
@mackyle have you had a chance to review my feedback? Also, as noted in #62, I plan to review all the TLS-related tickets together so that solutions will not conflict.

gstrauss:
@mackyle have you had a chance to review my feedback? Also, as noted in #62, I plan to review all the TLS-related tickets together so that solutions will not conflict.

gstrauss:
@mackyle have you had a chance to review my feedback? Also, as noted in #62, I plan to review all the TLS-related tickets together so that solutions will not conflict.

There are additional tickets which request other env vars.

src/response.c
```diff
@@ -136,16 +136,51 @@ static void https_add_ssl_entries(connection *con) {
...
+		(vr = SSL_get_verify_result(con->ssl)) != X509_V_OK
 		|| !(xs = SSL_get_peer_certificate(con->ssl))
 	) {
+		buffer_copy_string(ds_cv->value, (vr == X509_V_OK && !xs) ? "NONE" : "FAILED:bad verify result");
+		array_insert_unique(con->environment, (data_unset *)ds_cv);
 		return;
 	}
 
+	buffer_copy_string(ds_cv->value, "SUCCESS");
+	array_insert_unique(con->environment, (data_unset *)ds_cv);
```
gstrauss: above two lines should be part of the SSL_get_verify_result() condition block above to show that the outcome of the test sets SSL_CLIENT_VERIFY depending on the outcome of the condition test.

array_set_key_value() should be preferred over array_insert_unique() when it is desirable to replace an existing entry, if it exists, rather than having it handled by data_string insert_dup(), and perform a list append (combining the values with ", ")   (And yes, other code should be modified, too, but that should be handled in a separate pull request)

src/response.c
gstrauss: https_add_ssl_entries() is called near the start of the response handling, though it might be re-run if a request is restarted (e.g. via mod_rewrite), another reason why array_set_key_value() should be preferred in this code.

Additionally, this cut-n-paste of REMOTE_USER from the code below should be moved below the next condition block, as should the REMOTE_USER code in the block below.  Instead of duplicating the code, a simple array lookup can be done for (data_string *)array_get_element(con->environment, con->conf.ssl_verifyclient_username), and, if it exists, the value copied in REMOTE_USER, using array_set_key_value().

gstrauss:
The manpage for X509_NAME_oneline() says:
`The functions X509_NAME_oneline() and X509_NAME_print() are legacy functions which produce a non standard output form, they don't handle multi character fields and have various quirks and inconsistencies. Their use is strongly discouraged in new applications.`

Please consider using X509_NAME_print_ex() and propose a reasonable set of flags for a consistent and still-useful result.

PR #64, opened 2016-05-11
ssl: add support for ssl.cadn-file

mackyle:
If ssl.cadn-file is not set, fallback to ssl.ca-file.

The ssl.cadn-file option provides independent control of
the "certificate_authorities" field (see RFC 5246 section
7.4.4 Certificate Request) separate from the actual list
of trusted certificate authorities used for client
certificate verification.

It may be necessary to send a hint that includes the DN
of a non-root client CA in order to receive the correct
certificate from the client, but such a non-root CA really
does not belong in the trusted client root CA list.

Signed-off-by: Kyle J. McKay mackyle@gmail.com

### DIFF
--- a/src/base.h
+++ b/src/base.h
@@ -278,6 +278,7 @@ typedef struct {
 	/* server wide */
 	buffer *ssl_pemfile;
 	buffer *ssl_ca_file;
+	buffer *ssl_cadn_file;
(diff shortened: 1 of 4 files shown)


gstrauss:
x-ref: https://redmine.lighttpd.net/issues/2694

PR #65, opened 2016-05-25
Make the use of accept-filters conditional.

jsonn:
We had a recent discussion about bad interactions between accept-filters and misbehaving clients in NetBSD. The patch makes the use of accept filters conditional under an option, the default behavior doesn't t change.

### DIFF
--- a/src/configfile.c
+++ b/src/configfile.c
@@ -118,6 +118,7 @@ static int config_insert(server *srv) {
 		{ "server.http-parseopt-header-strict",NULL, T_CONFIG_BOOLEAN, T_CONFIG_SCOPE_SERVER     }, /* 72 */
 		{ "server.http-parseopt-host-strict",  NULL, T_CONFIG_BOOLEAN, T_CONFIG_SCOPE_SERVER     }, /* 73 */
 		{ "server.http-parseopt-host-normalize",NULL,T_CONFIG_BOOLEAN, T_CONFIG_SCOPE_SERVER     }, /* 74 */
+		{ "server.use-acceptfilter",           NULL, T_CONFIG_BOOLEAN, T_CONFIG_SCOPE_CONNECTION }, /* 75 */
 
 		{ "server.host",
 			"use server.bind instead",
@@ -185,6 +186,7 @@ static int config_insert(server *srv) {
 	cv[72].destination = &(srv->srvconf.http_header_strict);
 	cv[73].destination = &(srv->srvconf.http_host_strict);
 	cv[74].destination = &(srv->srvconf.http_host_normalize);
+	cv[75].destination = &(s->use_acceptfilter);
 
 	srv->config_storage = calloc(1, srv->config_context->used * sizeof(specific_config *));
 
@@ -221,6 +223,11 @@ static int config_insert(server *srv) {
 		s->use_ipv6      = 0;
 		s->set_v6only    = 1;
 		s->defer_accept  = 0;
+#ifdef SO_ACCEPTFILTER
+		s->use_acceptfilter = 1;
+#else
+		s->use_acceptfilter = 0;
+#endif
 #ifdef HAVE_LSTAT
 		s->follow_symlink = 1;
 #endif
--- a/src/base.h
+++ b/src/base.h
@@ -294,6 +294,7 @@ typedef struct {
 
 	unsigned short use_ipv6, set_v6only; /* set_v6only is only a temporary option */
 	unsigned short defer_accept;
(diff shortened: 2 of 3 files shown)


src/configfile.c
```diff
@@ -185,6 +186,7 @@ static int config_insert(server *srv) {
 	cv[72].destination = &(srv->srvconf.http_header_strict);
 	cv[73].destination = &(srv->srvconf.http_host_strict);
 	cv[74].destination = &(srv->srvconf.http_host_normalize);
+	cv[75].destination = &(s->use_acceptfilter);
```
gstrauss: Please move this line inside the loop immediately below, as it is T_CONFIG_SCOPE_CONNECTION.  (Sorry, the config option parsing code is more complex than I'd like.)

gstrauss:
Patch looks reasonable.

I added a comment inline in the diff for something that needs to be fixed.

Would you mind also changing use-acceptfilter to bsd-acceptfilter? (and use_acceptfilter to bsd_acceptfilter).  Thanks.

gstrauss:
Hmmm.  Giving this another look.  Why not re-use 'server.defer-accept' to enable/disable socket option TCP_DEFER_ACCEPT (Linux), or socket option SO_ACCEPTFILTER "httpready" (BSD)?  That would avoid adding another directive.

gstrauss:
Would the following work for you?  (Might need to make one additional change to enable s->defer_accept on BSD in order to preserve existing behavior of having "httpready" SO_ACCEPTFILTER enabled by default.  However, if that is not robust and not recommended by *BSD developers due to various client incompatibilities, then let's change the default, too.)

```
diff --git a/src/network.c b/src/network.c
index 19364ea..43e6af3 100644
--- a/src/network.c
+++ b/src/network.c
@@ -441,14 +441,13 @@ static int network_server_init(server *srv, buffer *host_token, specific_config
 
                goto error_free_socket;
 #endif
-#ifdef TCP_DEFER_ACCEPT
        } else if (s->defer_accept) {
+#ifdef TCP_DEFER_ACCEPT
                int v = s->defer_accept;
                if (-1 == setsockopt(srv_socket->fd, IPPROTO_TCP, TCP_DEFER_ACCEPT, &v, sizeof(v))) {
                        log_error_write(srv, __FILE__, __LINE__, "ss", "can't set TCP_DEFER_ACCEPT: ", strerror(errno));
                }
 #endif
-       } else {
 #ifdef SO_ACCEPTFILTER
                /* FreeBSD accf_http filter */
                struct accept_filter_arg afa;
```

jsonn:
Renaming to bsd_acceptfilter is something I'm ambivalent about. 

TCP_DEFER_ACCEPT is quite different and much more limited. I'm not sure if mixing the two things is really an improvement.

gstrauss:
Rather than getting side-tracked with platform-specific comparisons, let's focus first on the functional need addressed by this patch: you stated that there is a need to have an option for enabling/disabling "httpready" SO_ACCEPTFILTER.  I agree.

What is the general opinion of BSD core developers about enabling "httpready" SO_ACCEPTFILTER by default or not in lighttpd?  If there are good reasons to have the default be disabled, then please add that to the patch.  Thanks.

jsonn:
I think semantically, Linux's TCP_DEFER_ACCEPT and the *BSD's "dataready" SO_ACCEPTFILTER are equivalent. The "httpready" filter is more involved as it includes a partial HTTP parser. There are users that don't want that much parsing to happen in the kernel, especially on low-load machines it doesn't buy much.

src/configfile.c
```diff
@@ -221,6 +222,11 @@ static int config_insert(server *srv) {
 		s->use_ipv6      = 0;
 		s->set_v6only    = 1;
 		s->defer_accept  = 0;
+#ifdef SO_ACCEPTFILTER
+		s->use_acceptfilter = 1;
+#else
+		s->use_acceptfilter = 0;
+#endif
```
gstrauss: Since this is a BSD-specific flag, do the BSD core developers recommend that the default be enabled or disabled?  (Previously behavior in lighttpd was always enabled, so thank you for attempting to preserve that behavior.)

gstrauss:
@jsonn: still waiting for answer to question:
Since this is a BSD-specific flag, do the BSD core developers recommend that the default be enabled or disabled? (Previously behavior in lighttpd was always enabled, so thank you for attempting to preserve that behavior.)

jsonn:
I don't know of any such recommendation, in either direction. It is kind of a special purpose feature. There have been bugs in the filter in the past, but they have also been fixed.

PR #68, opened 2016-07-17
configure.ac: move inet_pton detection later, with inet_nto{a,p}.

fbrosson:
HAVE_INET_PTON was probably not being defined on Solaris.
While at it, also add detection for accept() in libnetwork for Haiku.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -136,7 +136,6 @@ dnl AC_FUNC_REALLOC
 AC_TYPE_SIGNAL
 AC_FUNC_STAT
 AC_FUNC_STRFTIME
-AC_CHECK_FUNCS([issetugid inet_pton])
 
 if test -z "$PKG_CONFIG"; then
   AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
@@ -283,6 +282,9 @@ dnl openssl on solaris needs -lsocket -lnsl
 AC_SEARCH_LIBS(socket,socket)
 AC_SEARCH_LIBS(gethostbyname,nsl socket)
 AC_SEARCH_LIBS(hstrerror,resolv)
+
+dnl On Haiku accept() and friends are in libnetwork
+AC_SEARCH_LIBS(accept,network)
 
 save_LIBS=$LIBS
 AC_SEARCH_LIBS(dlopen,dl,[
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -136,7 +136,7 @@ dnl AC_FUNC_REALLOC
 AC_TYPE_SIGNAL
 AC_FUNC_STAT
 AC_FUNC_STRFTIME
-AC_CHECK_FUNCS([issetugid inet_pton])
+AC_CHECK_FUNCS([issetugid])
```
gstrauss: would you move issetugid down with the others, too?

fbrosson: Sure. That's it.
BTW, should I change the commit message?

fbrosson: I'm updating the package for lighttpd 1.4.40 in Haiku.
The nightly builds of Haiku are available for 3 architectures: x86_gcc2, x86 and x86_64.
x86 and x86_64 use recent versions of gcc. But x86_gcc2 uses a slightly modified version of gcc-2.95.3, which means it does not accept most of what newer versions of gcc accept.

https://github.com/fbrosson/haikuports/tree/lighttpd/www-servers/lighttpd/patches
has the patches we need for lighttpd-1.4.40 to make it build on Haiku/x86_gcc2.
Do you think any of these two patches:
* [lighttpd-1.4.40-gcc2-mod_ssi-and-test_base64.patchset](https://github.com/fbrosson/haikuports/blob/lighttpd/www-servers/lighttpd/patches/lighttpd-1.4.40-gcc2-mod_ssi-and-test_base64.patchset)
* [lighttpd-1.4.40-gcc2-vector_config_weak.patchset](https://github.com/fbrosson/haikuports/blob/lighttpd/www-servers/lighttpd/patches/lighttpd-1.4.40-gcc2-vector_config_weak.patchset)

could be accepted in lighttpd?
Please feel free to answer no, since we are already used to having plenty of gcc2 patches for many programs ;-)
BTW, I'll dig why I added a **`(unsigned char)`** cast in some chunks but not all.
Thanks!

gstrauss: FYI: line note comments might get lost after you push a commit which replaces the commit with the line notes.  Also, you can send off-topic, or tangentially-related questions, such as this, to me via email (see https://github.com/gstrauss/).

To answer your question: sorry, as is demonstrated in lighttpd-1.4.40-gcc2-std-gnu9x.patchset, lighttpd is compiled towards -std=gnu99.  I am considering changing that to -std=gnu11.  C99 was published over 17 years ago.  C11 was published almost 5 years ago.  Let that sink in for a moment. :)  I took a quick look at your patches and had all but forgotten about "not a static initializer" errors.  Also, I don't recall ever using a preprocessor that was unable to support token pasting (##).  While I am sure you have a reason for using gcc2, have you considered porting whatever needs to be ported into clang, if licensing with gcc is a sticking point?  We probably should continue this off-topic conversation offline.

fbrosson: No problem. Thanks for your reply.
Regarding C99, Haiku/x86_gcc2 still uses it because the core developers chose to maintain binary compatibility with BeOS. But that will be dropped someday.
Anyway, thank you very much for the **`AC_SEARCH_LIBS(accept,network)`**. This means lighttpd 1.4.41 won't need any patch on Haiku/x86{,_64} which are already using gcc 5.3 ;-)

PR #69, opened 2016-07-27
allow setting test port via env var

glensc:
taken from https://github.com/pld-linux/lighttpd/commit/5f5b97ee73ba4151b4d707dcdd6859edd2e18c1f

### DIFF
--- a/tests/LightyTest.pm
+++ b/tests/LightyTest.pm
@@ -78,7 +78,7 @@ sub new {
 		$self->{MODULES_PATH} = $self->{BASEDIR}.'/build';
 	}
 	$self->{LIGHTTPD_PATH} = $self->{BINDIR}.'/lighttpd';
-	$self->{PORT} = 2048;
+	$self->{PORT} = $ENV{LIGHTTPD_TEST_PORT} ? int($ENV{LIGHTTPD_TEST_PORT}) : 2048;
 
 	my ($name, $aliases, $addrtype, $net) = gethostbyaddr(inet_aton("127.0.0.1"), AF_INET);
 


gstrauss:
* In keeping with labels familiar to CGI environment, please consider naming the variable LIGHTTPD_SERVER_PORT instead of LIGHTTPD_TEST_PORT.  Then, LIGHTTPD_BACKEND_PORT could specify the port used by additional server test instances, such as in the proxy tests.
* tests/*.conf specifies server.port = 2048, so those would need to be modified to be env.PORT and added to your patch
* tests/proxy.conf also specifies server.port = 2050 for the proxy instance and would need to be exported into the test environment to be used in proxy.conf as env.LIGHTTPD_BACKEND_PORT

gstrauss:
@glensc: have you had a chance to review my feedback?

glensc:
@gstrauss yes. i will do it. just not sure when

gstrauss:
Going to close this for now.  Please re-open when you have a chance to submit another patch for review and I'll take another look.

`$ grep port tests/*.conf`
Currently, tests/*.conf use ports 1026, 1048, 2048, 2050, 10000, 20000 for the servers, backend proxies, and other backends (e.g. fastcgi)  (The use of port 1048 in tests/fastcgi-13.conf should probably be changed to the same as port 1026)

glensc:
This got improved in 1.4.59:
- [tests] use ephemeral ports in tests

ref:
- https://redmine.lighttpd.net/projects/lighttpd/wiki/Release-1_4_59/

PR #70, opened 2016-08-13
[mod_fastcgi] FastCGI Authorizer support with FastCGI Responders (fixes #321, #322)

ck-ws:
As discussed in http://redmine.lighttpd.net/issues/322, I reworked the patch from scratch based on git-master.
I also changed the test FastCGI authorizer and responder as well as the test case to test the new behavior.

Changes from old patch:
- I save the fcgi-exts in in FastCGI-Authorizer env var instead of just a flag, so it's theoretically possible to configure and use more than one authorizer
- as advised by you, array search is now outside of the loop (that was stupid of me ;) ) and buffer->used isn't needed any longer

I tried to not hurt the performance - if you see place for improvements, please comment on the diff :)

TODO:
- [ ] Rewrite extension check to use |ext|

Regards,
Christoph (ckreutzer)

### DIFF
--- a/src/mod_fastcgi.c
+++ b/src/mod_fastcgi.c
@@ -1511,11 +1511,6 @@ SETDEFAULTS_FUNC(mod_fastcgi_set_defaults) {
 							host->mode = FCGI_RESPONDER;
 						} else if (strcmp(fcgi_mode->ptr, "authorizer") == 0) {
 							host->mode = FCGI_AUTHORIZER;
-							if (buffer_string_is_empty(host->docroot)) {
-								log_error_write(srv, __FILE__, __LINE__, "s",
-										"ERROR: docroot is required for authorizer mode.");
-								goto error;
-							}
 						} else {
 							log_error_write(srv, __FILE__, __LINE__, "sbs",
 									"WARNING: unknown fastcgi mode:",
@@ -2397,6 +2392,19 @@ range_success: ;
 		default:
 			break;
 		}
+
+		if (host->mode == FCGI_AUTHORIZER &&
+		    key_len > 9 &&
+		    0 == strncasecmp(key, CONST_STR_LEN("Variable-"))) {
+			data_string *ds;
+			if (NULL == (ds = (data_string *)array_get_unused_element(con->environment, TYPE_STRING))) {
+				ds = data_response_init();
+			}
+			buffer_copy_string_len(ds->key, key + 9, key_len - 9);
+			buffer_copy_string(ds->value, value);
+
+			array_insert_unique(con->environment, (data_unset *)ds);
+		}
 	}
 
 	if (have_sendfile2) {
@@ -3269,6 +3277,7 @@ static handler_t fcgi_recv_response(server *srv, handler_ctx *hctx) {
 	plugin_data *p    = hctx->plugin_data;
 
 	fcgi_proc *proc   = hctx->proc;
+	fcgi_extension *ext = hctx->ext;
 	fcgi_extension_host *host= hctx->host;
 
 		switch (fcgi_demux_response(srv, hctx)) {
@@ -3285,16 +3294,46 @@ static handler_t fcgi_recv_response(server *srv, handler_ctx *hctx) {
 				 * now to handle authorized request.
 				 */
 
-				buffer_copy_buffer(con->physical.doc_root, host->docroot);
-				buffer_copy_buffer(con->physical.basedir, host->docroot);
+				if (!buffer_is_empty(host->docroot)) {
+					buffer_copy_buffer(con->physical.doc_root, host->docroot);
+					buffer_copy_buffer(con->physical.basedir, host->docroot);
 
-				buffer_copy_buffer(con->physical.path, host->docroot);
-				buffer_append_string_buffer(con->physical.path, con->uri.path);
+					buffer_copy_buffer(con->physical.path, host->docroot);
+					buffer_append_string_buffer(con->physical.path, con->uri.path);
 
-				con->mode = DIRECT;/*(avoid changing con->state, con->http_status)*/
-				fcgi_connection_close(srv, hctx);
-				con->http_status = 0;
-				con->file_started = 1; /* fcgi_extension won't touch the request afterwards */
+					con->mode = DIRECT;/*(avoid changing con->state, con->http_status)*/
+					fcgi_connection_close(srv, hctx);
+					con->http_status = 0;
+					con->file_started = 1; /* fcgi_extension won't touch the request afterwards */
+				} else {
+					/* docroot was not set - restart the request so other handlers can catch it */
+					fcgi_connection_close(srv, hctx);
+
+					connection_response_reset(srv, con); /*(includes con->http_status = 0)*/
+					con->mode = DIRECT;
+
+					/* user was authorized, set the FastCGI authorizer flag */
+					data_string *authorizer_env = NULL;
+					if (NULL == (authorizer_env = (data_string *)array_get_unused_element(con->environment, TYPE_STRING))) {
+						authorizer_env = data_string_init();
+					}
+					buffer_copy_string(authorizer_env->key, "FastCGI-Authorizer");
+					buffer_copy_buffer(authorizer_env->value, ext->key);
+					array_insert_unique(con->environment, (data_unset *)authorizer_env);
+
+					/* don't do more than 6 loops here, that normally shouldn't happen */
+					if (++con->loops_per_request > 5) {
+						authorizer_env = (data_string *)array_get_element(con->environment, "FastCGI-Authorizer");
+						log_error_write(srv, __FILE__, __LINE__, "sb", "too many authorizer loops while processing request:", con->request.orig_uri);
+						if (NULL != authorizer_env) {
+							log_error_write(srv, __FILE__, __LINE__, "sb", "FastCGI-Authorizer: ", authorizer_env->value);
+						}
+						con->http_status = 500; /* Internal Server Error */
+						return HANDLER_FINISHED;
+					}
+
+					return HANDLER_COMEBACK;
+				}
 			} else {
 				/* we are done */
 				fcgi_connection_close(srv, hctx);
@@ -3489,6 +3528,8 @@ static handler_t fcgi_check_extension(server *srv, connection *con, void *p_d, i
 	fcgi_extension *extension = NULL;
 	fcgi_extension_host *host = NULL;
 
+	data_string *authorizer_env = (data_string *)array_get_element(con->environment, "FastCGI-Authorizer");
+
 	if (con->mode != DIRECT) return HANDLER_GO_ON;
 
 	/* Possibly, we processed already this request */
@@ -3551,6 +3592,27 @@ static handler_t fcgi_check_extension(server *srv, connection *con, void *p_d, i
 			fcgi_extension *ext = p->conf.exts->exts[k];
 
 			if (buffer_is_empty(ext->key)) continue;
+
+			/* do not reuse same authorizer, skip if it's in FastCGI-Authorizer env var */
+			if (NULL != authorizer_env) {
+				if (0 == strcmp(authorizer_env->value->ptr, ext->key->ptr)) continue;
+
+				buffer *ext_leading = buffer_init_string(", ");
+				buffer_append_string_buffer(ext_leading, ext->key);
+
+				buffer *ext_trailing = buffer_init_buffer(ext->key);
+				buffer_append_string(ext_trailing, ", ");
+
+				if (NULL != strstr(authorizer_env->value->ptr, ext_leading->ptr) ||
+				    NULL != strstr(authorizer_env->value->ptr, ext_trailing->ptr)) {
+					buffer_free(ext_leading);
+					buffer_free(ext_trailing);
+					continue;
+				} else {
+					buffer_free(ext_leading);
+					buffer_free(ext_trailing);
+				}
+			}
 
 			ct_len = buffer_string_length(ext->key);
 
--- a/tests/fastcgi-auth.conf
+++ b/tests/fastcgi-auth.conf
@@ -68,13 +68,21 @@ compress.filetype = (
 fastcgi.debug = 0
 fastcgi.server = (
 	"/" => (
-			"grisu" => (
+		"grisu" => (
 			"host" => "127.0.0.1",
 			"port" => 20000,
 			"bin-path" => env.SRCDIR + "/fcgi-auth",
 			"mode" => "authorizer",
-			"docroot" => env.SRCDIR + "/tmp/lighttpd/servers/www.example.org/pages/",
 			"check-local" => "disable",
+		),
+	),
+	".fcgi" => (
+		"grisu2" => (
+			"host" => "127.0.0.1",
+			"port" => 10000,
+			"bin-path" => env.SRCDIR + "/fcgi-responder",
+			"check-local" => "disable",
+			"max-procs" => 1,
 		),
 	),
 )
--- a/tests/fcgi-auth.c
+++ b/tests/fcgi-auth.c
@@ -19,8 +19,11 @@ int main (void) {
 		printf("Content-type: text/html\r\n");
 
 		if (((p = getenv("QUERY_STRING")) == NULL) ||
-		    strcmp(p, "ok") != 0) {
+		    (strcmp(p, "ok") != 0 && strcmp(p, "var") != 0)) {
 			printf("Status: 403 Forbidden\r\n\r\n");
+		} else if (((p = getenv("QUERY_STRING")) != NULL) || strcmp(p, "var") == 0) {
+			printf("Variable-LIGHTYTEST: LighttpdTestContent\r\n");
+			printf("\r\n");
 		} else {
 			printf("\r\n");
 			/* default Status is 200 - allow access */
--- a/tests/fcgi-auth.c
+++ b/tests/fcgi-auth.c
@@ -11,25 +11,18 @@
 #include <string.h>
 
 int main (void) {
-	char* p;
-
 	while (FCGI_Accept() >= 0) {
 		/* wait for fastcgi authorizer request */
+		char* p = getenv("QUERY_STRING");
 
-		printf("Content-type: text/html\r\n");
-
-		if (((p = getenv("QUERY_STRING")) == NULL) ||
-		    (strcmp(p, "ok") != 0 && strcmp(p, "var") != 0)) {
-			printf("Status: 403 Forbidden\r\n\r\n");
-		} else if (((p = getenv("QUERY_STRING")) != NULL) && strcmp(p, "var") == 0) {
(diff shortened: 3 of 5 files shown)


tests/fcgi-auth.c
```diff
@@ -19,8 +19,11 @@ int main (void) {
 		printf("Content-type: text/html\r\n");
 
 		if (((p = getenv("QUERY_STRING")) == NULL) ||
-		    strcmp(p, "ok") != 0) {
+		    (strcmp(p, "ok") != 0 && strcmp(p, "var") != 0)) {
 			printf("Status: 403 Forbidden\r\n\r\n");
+		} else if (((p = getenv("QUERY_STRING")) != NULL) || strcmp(p, "var") == 0) {
```
gstrauss: This test looks incorrect.  Should it be "&&" instead of "||" ?

tests/fastcgi-auth.conf
```diff
@@ -68,15 +68,23 @@ compress.filetype = (
 fastcgi.debug = 0
 fastcgi.server = (
 	"/" => (
-			"grisu" => (
+		"grisu" => (
 			"host" => "127.0.0.1",
 			"port" => 20000,
 			"bin-path" => env.SRCDIR + "/fcgi-auth",
 			"mode" => "authorizer",
-			"docroot" => env.SRCDIR + "/tmp/lighttpd/servers/www.example.org/pages/",
```
gstrauss: Does this change affect other tests?  (I have not looked closely)

src/mod_fastcgi.c
```diff
@@ -3552,6 +3593,27 @@ static handler_t fcgi_check_extension(server *srv, connection *con, void *p_d, i
 
 			if (buffer_is_empty(ext->key)) continue;
 
+			/* do not reuse same authorizer, skip if it's in FastCGI-Authorizer env var */
```
gstrauss: What is the use case here?  con->loops_per_request test elsewhere will prevent infinite loops.
Should the 'continue' be 'break'?  Should the check be done once below the loop, only if there is a match?

Also, the implementation below only be a few lines of code, but it could probably be more efficient if you also modify the code that inserts FastCGI-Authorizer into the environment.  If you start FastCGI-Authorizer with '|' and modify the existing value and always append 'ext|', then you can strstr for '|ext|'.

BTW, I am still mulling over the choice of 'FastCGI-Authorizer' env var name.  It will leak into CGI, FastCGI, SCGI, and SSI environments, since lighttpd does not have the equivalent to Apache module 'notes', which do not leak into subprocess environments.

tests/fcgi-auth.c
ck-ws: Yeah, that's correct. Don't know why I ORed that, either.

gstrauss:
I'll try to find some time soon to reload context and dig into the core of the patch to be able to comment further.  Thanks for your effort!

tests/fastcgi-auth.conf
ck-ws: In short: No.
The configuration is only used for the FastCGI authorizer tests. These were previously 3, and are now 5. The requested files (`/index.html` and `/expire/access.txt`) are properly delivered, because `server.document-root` is the same as `docroot` and the request is now reloaded, instead of delivering directly out of `docroot` (which is also possible, I haven't removed the configuration option).

All the tests pass fine also:
`````
./mod-fastcgi.t ....... ok     
[...]
All tests successful.
Files=25, Tests=403,  6 wallclock secs ( 0.13 usr  0.02 sys +  1.27 cusr  0.17 csys =  1.59 CPU)
Result: PASS
PASS: run-tests.pl
cleaning up                             PASS: cleanup.sh
==================
All 3 tests passed
==================
`````

File: src/mod_fastcgi.c (hunk 3 above)
Comment (ck-ws) [inline src/mod_fastcgi.c]: > What is the use case here? con->loops_per_request test elsewhere will prevent infinite loops.

`con->loops_per_request` will prevent infinite loops, but cancels the request with a 5xx error (in case there is something wrong in the code or the config of authorizers).

> Should the 'continue' be 'break'? Should the check be done once below the loop, only if there is a match?

No, that's intended. Usually, there is only one authorizer, but IMHO it is possible that there is more than one.

The second time the request is handled, we are iterating over all the configured extensions. Without checking the env var, the request would be rescheduled to the same authorizer, which would result in an infinite loop. If we would do a check based on if a flag is set and the extension is for an authorizer (like in the old patch), only one authorizer could be used. Checking the extensions like we do here, we can "jump" over the already used ones (which are listed in the env var) by using `continue`, and select the next one in the array which fullfills the requirements for processing the request. Concluded, this allows multiple authorizers and the last handling of a responder.

> Also, the implementation below only be a few lines of code, but it could probably be more efficient if you also modify the code that inserts FastCGI-Authorizer into the environment. If you start FastCGI-Authorizer with '|' and modify the existing value and always append 'ext|', then you can strstr for '|ext|'.

Wow, that would be much simpler. I only thought about classical header formats, and the multiple value handling came in handy, as it was already there.
I would implement it like this, but I will wait, if you maybe come up with a better solution instead of that env var.
Just as a note: In the case only one authorizer is configured (which should be the most common), only the most simple `strcmp` is used :) So it usually should be efficient, but the implementation of your idea would be much cleaner.

> BTW, I am still mulling over the choice of 'FastCGI-Authorizer' env var name. It will leak into CGI, FastCGI, SCGI, and SSI environments, since lighttpd does not have the equivalent to Apache module 'notes', which do not leak into subprocess environments.

Yeah, I thought the same. Unfornately, I didn't found anything more suitable in `src/base.h#L387`. I don't think it is much of a problem, because the application only can see that there was an authorizer handling the request and for which prefix/extension it was configured.
Anyway, if you would add sth. like a `notes` array to the `connection` struct, I would rewrite the patch to use it. But I'm not sure if that is overkill, as I don't know how that would influence the overall performance, since it will be unused for most of the requests.

Or we could implement it like the HTTP_PROXY filter in `fcgi_env_add_request_headers`, so we don't forward our env var to the processes. But that seems a bit hacky.

**Comment by gstrauss** [comment]
Attempting to get myself back up to speed on this one.

I think that the chunks of code related to potentially allowing multiple authorizers should be removed, and possibly submitted in a separate pull request.  Sorry, I am still not seeing how the extension of the request URL will change between calls to connection_response_reset(), leading to a different authorizer being chosen on a subsequent pass, unless one is an anchored path and the other is an extension.  (And if that is the case, why restart the request?  We might consider walking the array looking for subsequent matching FastCGI authorizers, instead of restarting the request.)

Separating out the "multiple authorizers" code can let us focus on the actual thrust of this pull request, which is to allow for FastCGI responders, and to honor Variable-* response headers from the FastCGI authorizer.

It appears that you are using the non-existence of host->docroot as a flag to indicate that the request should be restarted after authorization, and presence of host->docroot meaning that a static file should be delivered after authorization (the current behavior).  Is there a better way to do this? (maybe, maybe not)  Should the request always be restarted, and mod_fastcgi recognize that the request was already authorized, and just let things pass through so that mod_staticfile handles the request?

If we separate out the "multiple authorizers" code, how shall we distinguish a URL extension for FastCGI authorizer from the same URL extension for FastCGI responder?  Should they be stored in separate arrays?

**Comment by ck-ws** [comment]
> Sorry, I am still not seeing how the extension of the request URL will change between calls to connection_response_reset(), leading to a different authorizer being chosen on a subsequent pass, unless one is an anchored path and the other is an extension.

Maybe extension is the wrong word here - I used it, since it is used in the source code. I would say, it is more common that for an authorizer a prefix match will be used. Even the prefix in the request can't change, but think of the following configuration:
````
/ => Authorizer
/secured => Authorizer
.fcgi => Responder
````
Now you have a request to `/secured/foobar.fcgi`. Which authorizer should be chosen? IMO, both. But we could say only the second, since it is more specific. Or the first, because it is the first (which was the bevavior in the old patch).

> (And if that is the case, why restart the request? We might consider walking the array looking for subsequent matching FastCGI authorizers, instead of restarting the request.)

Sorry, can't follow you here. Can you link to the code, please? :) In my perception, we use the first "extension" (or prefix) that matches, may it be an authorizer or responder. If it is an authorizer, the request gets rescheduled and comes there again. Then, we `continue` over the previous authorizer and may choose another authorizer or "finish" the request when selecting a responder.

I don't know how you would apply more than one authorizer without restarting the request in between.

If you like, I can exclude the handling of the multiple authorizers - for me one is perfectly enough. I don't know if there is so. out there using more than one. When writing this patch it came to my eyes that it would be possible to do it generically, so I did. Maybe we are just misunderstanding each other.

------

> It appears that you are using the non-existence of host->docroot as a flag to indicate that the request should be restarted after authorization, and presence of host->docroot meaning that a static file should be delivered after authorization (the current behavior). Is there a better way to do this? (maybe, maybe not) Should the request always be restarted, and mod_fastcgi recognize that the request was already authorized, and just let things pass through so that mod_staticfile handles the request?

I tried not to remove the functionality that was already there. For me it would fine if mod_staticfile would handle the request afterwards (it already does if host->docroot is missing). But since this is a breaking change (currently, only the handling using host->docroot is supported, and that doesn't need to be same path as server.document-root), I think it would be better to leave it for backward compatibility.
Maybe it could be removed for an upcoming 1.5 (not the original 1.5, more like a 1.4 with breaking changes) or for 2.0 (but I think there isn't any authorizer support at all, atm).

> If we separate out the "multiple authorizers" code, how shall we distinguish a URL extension for FastCGI authorizer from the same URL extension for FastCGI responder? Should they be stored in separate arrays?

If I remember correctly, it is currently impossible to configure the same extension more than once, so it can't be an authorizer and responder at the same time. Why would it be useful? I don't think it is necessary to split the arrays apart.

Thanks again for checking back :) Have you decided to replace the Authorizer-Header using sth. like `notes` (my comment from 2016/08/19)?

File: tests/fcgi-auth.c [hunk 4]
```diff
@@ -21,7 +21,7 @@ int main (void) {
 		if (((p = getenv("QUERY_STRING")) == NULL) ||
 		    (strcmp(p, "ok") != 0 && strcmp(p, "var") != 0)) {
 			printf("Status: 403 Forbidden\r\n\r\n");
-		} else if (((p = getenv("QUERY_STRING")) != NULL) || strcmp(p, "var") == 0) {
+		} else if (((p = getenv("QUERY_STRING")) != NULL) && strcmp(p, "var") == 0) {
```
gstrauss: No need to repeat (p = getenv("QUERY_STRING")) here.  It was done above.  Still need to check (p != NULL).  Sorry that I hadn't mentioned this in prior comment.  (BTW, feel free to rebase and force-push changes to this dev branch.)

I'd move the char* p into the loop, too, and remote the unnecessary Content-Type response header (not required for FastCGI authorizer).  (Untested code typed directly into web browser:)
<pre>
                /* wait for fastcgi authorizer request */

                /* Status: 200 OK to allow access is implied
                 * if Status header is not included in response */
                char *p = getenv("QUERY_STRING");
                if (p != NULL && 0 == strcmp(p, "ok")) {
                        printf("\r\n");
                } else if (p != NULL && 0 == strcmp(p, "var")) {
                        printf("Variable-X-LIGHTTPD-FCGI-AUTH: LighttpdTestContent\r\n\r\n");
                } else {
                        printf("Status: 403 Forbidden\r\n\r\n");
                }
</pre>
I also renamed LIGHTYTEST to X-LIGHTTPD-FCGI-AUTH, though doing so here will mean updating other tests in this pull request which looking for that header.

gstrauss:
FYI, I am working on some code to give examples what I meant and to address some of the questions you posed in your most recent comments last week.  I'll post a dev branch of code tomorrow evening.

tests/fcgi-responder.c
```diff
@@ -44,6 +44,10 @@ int main (void) {
 			printf("%s", getenv("PATH_INFO"));
 		} else if (0 == strcmp(p, "script_name")) {
 			printf("%s", getenv("SCRIPT_NAME"));
+		} else if (0 == strcmp(p, "var")) {
+			char *test = getenv("LIGHTYTEST");
+			if (test == NULL) test = "(no value)";
+			printf("%s", test);
```
ck-ws: I added a new commit to this branch fixing these issues.
You first commented that authorizer doesn't need the Content-Type header, so I removed it. All tests run fine. Or should I add it again, as you did?

gstrauss: I don't see why the Content-Type is needed, or the subsequent body since it should be discarded for 200 OK (but should be preserved for the 403 Forbidden).  I added it back when I noticed that a response body was being sent unconditionally.

I don't have a strong preference for or against its presence, as long as the tests pass.

gstrauss:
Found a couple more issues.

Since FastCGI authorizer previously only authorized access to a static file, this was not previously an issue, but now that the request is restarted, mod_fastcgi must not read request body (e.g. from POST) until after handling FastCGI authorizer(s).

<pre>
@@ -3233,9 +3241,16 @@ SUBREQUEST_FUNC(mod_fastcgi_handle_subrequest) {
                }
        }
 
-       if (0 == hctx->wb->bytes_in
-           ? con->state == CON_STATE_READ_POST
-           : hctx->wb->bytes_in < hctx->wb_reqlen) {
+       /* (do not receive request body before FCGI_AUTHORIZER has run or else
+        *  the request body is discarded with fcgi_connection_close() of the
+        *  FastCGI Authorizer)
+        * (hctx->host might not be assigned yet, so check first host in list
+        *  of potential hosts (hctx->ext->hosts[0]->mode != FCGI_AUTHORIZER))*/
+
+       if (hctx->ext->hosts[0]->mode != FCGI_AUTHORIZER
+           && (0 == hctx->wb->bytes_in
+               ? con->state == CON_STATE_READ_POST
+               : hctx->wb->bytes_in < hctx->wb_reqlen)) {
                /*(64k - 4k to attempt to avoid temporary files
                 * in conjunction with FDEVENT_STREAM_REQUEST_BUFMIN)*/
                if (hctx->wb->bytes_in - hctx->wb->bytes_out > 65536 - 4096
</pre>

Second issue is more vexing.  fastcgi.map-extensions was added to mod_fastcgi back in 2006.  fastcgi.map-extensions, as written, takes precedence over the order of the list of paths/extensions, searching for matching suffixes before processing the list of paths/extensions in the order given in the config.  (Your patch currently does not handle matching authorizer due to fastcgi.map-extensions and will result in spinning. sending the request back to the fastcgi authorizer repeatedly.)

Changing the behavior of mod_fastcgi to search the lists of paths/extensions in order given might break someone's existing config if they use fastcgi.map-extensions and also have paths in the paths/extensions list which would match prior to the mapped extension.  I have not decided what to do here yet.  One option is to considering this a bug, fix it, and note in release notes that some existing configs might break, though I would really like to avoid that.  Another option is to note in the documentation the niggling behavior and to keep it as-is, meaning that the request restart feature in your patch (which is used in a way to send request to a different FastCGI handler) will not be available in conjunction with fastcgi.map-extensions unless the config file is written just-so where the mapping extension is the authorizer and a later path match sends the request to the fastcgi handler -- and not the other way around, despite any ordering of paths/extensions in the list in the config file.  This second option is more likely, even though more complex, since it preserves historical behavior and will not break existing configs.

Your patch will still need to be updated to skip an already-handled extension which matched in the processing of fastcgi.map-extensions.  FYI: I have spent a bit of time on this, but not sure if I'll be ready to post my branch tonight.  You might want to wait until I post it (today or tomorrow) to see what I do here.

ck-ws:
Thanks for your participation here. This is much more complex than I thought ;)
I would also opt for the non-breaking path - even if it's more complex. I would like to avoid breaking existing configurations, too.

gstrauss:
I pushed a test branch which rewrites the patch.
$ git clone https://git.lighttpd.net/lighttpd/lighttpd1.4.git
and please see the top two commits on my devel branch personal/gstrauss/master
(note that is a development branch and I rewrite history)  It contains a few patches past current lighttpd git master.

The merge-base of your branch was old and missing some important fixes (unrelated to your patch, but occurring in mod_fastcgi.c with overlapping lines)

I removed FastCGI-Authorizer from con->environment and replaced it with code which stores state in the per-request config array containing per-module module config for that request (con->plugin_ctx[p->id]) and modified the cleanup of that to keep state of whether or not a FastCGI authorizer had been run.  **Note** that this does not support multiple authorizers (though it could if more state was kept and cleaned up).

I cherry-picked your changes that we have discussed regarding the test framework.  (Thank you for writing tests!)

Regarding the issues I raised in my last comment, I have not modified the historical interaction between fastcgi.map-extensions and linear processing of the path/extension list, but should probably add some documentation in conjunction with this feature.

Also, I am still thinking about making a new config flag for fastcgi.server, rather than overloading the existence of docroot param, since a remote fastcgi server might need that information for its auth processing.

gstrauss:
FYI: current behavior of mod_fastcgi, which requires docroot param be set for FCGI_AUTHORIZER, is to modify SCRIPT_FILENAME and DOCUMENT_ROOT based on that parameter.  Overloading docroot as a flag for restarting the request is not a good idea since it removes the ability to transmit a custom docroot to a remote FastCGI backend when also wanting to restart the request once it has been authorized.

gstrauss:
With my changes to keep FCGI_AUTHORIZER state in per-request config for mod_fastcgi, I modified FCGI_AUTHORIZER handling to always restart the request if auth succeeds.

If fastcgi.server docroot param is set, the physical path is translated and preserved when the request is restarted, skipping earlier processing hooks (uri_raw, uri_clean, docroot, physical hooks, since path translation has already occurred and URL has not changed) and continuing on to mod_staticfile (with the option for dynamic handlers such as mod_fastcgi to handle the request instead).  If docroot param is not set, the request is restarted and all hooks are run.

I forced pushed to update personal/gstrauss/master branch in https://git.lighttpd.net/lighttpd/lighttpd1.4.git

gstrauss:
I updated personal/gstrauss/master branch in https://git.lighttpd.net/lighttpd/lighttpd1.4.git

mod_fastcgi now supports configuring FCGI_AUTHORIZER and FCGI_RESPONDER for same path-prefix or extension.

While the patch preserves what I consider unfortunate/broken behavior of fastcgi.map-extensions taking precedence over the list of paths/extensions for the fastcgi.server, being able to specify both an FCGI_AUTHORIZER and FCGI_RESPONDER for the same path/extension should minimize those affected by fastcgi.map-extensions and trying to use this new feature (FastCGI authorizer and responder support for the same request).

Does this patch set work for you?  It works for me and I plan to commit it to lighttpd git master in a couple days.

Thanks for your work on the initial patch set.

ck-ws:
Hi Glenn,
Thanks for your hard work on this - I attempted to do my best, but I think you know the internals of lighttpd much better than I.
I will check your implementation in the next days in my test and production environment and give you feedback. But since the tests pass, I think it will be fine.

ck-ws:
Hi Glenn,
you already merged it to master, but as promised:
I had the chance to test your implementation and it works fine in my environment.

Thanks again!

gstrauss:
Thanks for your testing and help!  The purpose of merging to master is to hopefully give others more time to test out the feature before the next release (even if that testing does not happen often enough in practice).

Thanks again.

PR #71, opened 2016-10-20
build: use CC_FOR_BUILD for lemon when cross-compiling

mtdcr:
This allows to generate some necessary files when cross-compiling from Git. I think release tarballs already contain these files, but I didn't double-check.

Note that source code compiled for the build machine shouldn't use config.h, because config.h matches the host machine's environment. That's why the #ifdefs in lemon.c were removed. inttypes.h is available since C99. Also note that `HAVE_CONFIG_H` now won't get defined for lemon, so first.h doesn't include config.h anymore.

`#include <stdint.h>` was removed from lemon.c, because no types declared in stdint.h are used in this file. FWIW, inttypes.h includes stdint.h anyway.

Regards,
Andreas

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,6 +63,7 @@ AC_DEFUN([TRY_LDFLAGS],
 dnl Checks for programs.
 AC_PROG_CC
 AM_PROG_CC_C_O
+AX_PROG_CC_FOR_BUILD
 AC_PROG_LD
 AC_PROG_INSTALL
 AC_PROG_AWK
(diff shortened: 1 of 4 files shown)


gstrauss:
Thanks for the patch!

BTW, lemon.c includes stdint.h or inttypes.h only for (intptr_t) definition.  inttypes.h has been around since C90, so much longer than stdint.h (C99).  Therefore, your change to lemon.c looks fine.  I might also remove the include of first.h since lemon.c is standalone, and lemon.c could just as easily define _GNU_SOURCE at the top.

gstrauss:
Merged patch using 'git am'  Not sure why github did not properly notice.

Thanks for the patch!

PR #72, opened 2016-11-02
fix systemd setting same as lighttpd2.

sugilog:
Hi.

I fixed systemd setting file in `docs/systemd`.
I need `reload` command when `logrotate` lighttpd access log.
After `logrotate`, need to send signal to change file descriptor of log file.

### DIFF
--- a/doc/systemd/lighttpd.service
+++ b/doc/systemd/lighttpd.service
@@ -3,8 +3,11 @@ Description=Lighttpd Daemon
 After=network.target
 
 [Service]
+Type=simple
 ExecStartPre=/usr/sbin/lighttpd -t -f /etc/lighttpd/lighttpd.conf
 ExecStart=/usr/sbin/lighttpd -D -f /etc/lighttpd/lighttpd.conf
+ExecReload=/bin/kill -HUP $MAINPID
+Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target


gstrauss:
>+Type=simple

(the default if neither Type= nor BusName=, but ExecStart= are specified)

> ExecStartPre=/usr/sbin/lighttpd -t -f /etc/lighttpd/lighttpd.conf

Would be better as ExecStartPre=/usr/sbin/lighttpd -tt -f /etc/lighttpd/lighttpd.conf

>+ExecReload=/bin/kill -HUP $MAINPID

Unfortunately, the HUP signal does not cause lighttpd 1.4.x to reload its configuration, which is the intention of systemd ExecReload=....

One of the secondary effects of reloading configs in web servers is typically to reopen log files, and you are correct that sending lighttpd 1.4.x a HUP signal causes mod_accesslog to reopen the accesslog files.  However, since HUP does not cause lighttpd 1.4.x to reload its configuration, I can not accept this change, as it does not perform all the expected behaviors.

If you want to rotate lighttpd logs with systemd, I suggest using a piped logger (e.g. cronolog) in the lighttpd.conf, rather than flat access log files, and then you can have systemd send a signal, if appropriate, to the piped logger when rotating logs.  You are, of course, also welcome to use ExecReload=/bin/kill -HUP $MAINPID in your own systemd/lighttpd.service, since you understand the limitations.

>+Restart=on-failure

This Restart=on-failure be a good addition to  doc/systemd/lighttpd.service

Closing this pull request since its primary purpose (signal for log rotation) can not be accepted into lighttpd 1.4.x branch at present due to issues described above.

sugilog:
Thank you. I will check comment above.

PR #73, opened 2016-11-11
[config] suggest inclusion of mod_geoip... before mod_ssi.

fbrosson:
* In modules.conf, mod_geoip needs to be loaded before mod_ssi, otherwise GeoIP vars won't be available to SSI pages.
* In geoip.conf suggest GeoLiteCity.dat instead of GeoCityLite.dat.

### DIFF
--- a/doc/config/conf.d/geoip.conf
+++ b/doc/config/conf.d/geoip.conf
@@ -16,7 +16,7 @@ server.modules += ( "mod_geoip" )
 ## mod_geoip will determine the database type automatically so if you
 ## enter GeoCity databse path it will load GeoCity Env.
 ##
-#geoip.db-filename = "/path/to/GeoCityLite.dat"
+#geoip.db-filename = "/path/to/GeoLiteCity.dat"
 
 ##
 ## If enabled, mod_geoip will load the database binary file to memory
--- a/doc/config/modules.conf
+++ b/doc/config/modules.conf
@@ -59,6 +59,11 @@ server.modules = (
 ##
 
 ##
+## mod_geoip
+##
+#include "conf.d/geoip.conf"
+
+##
 ## mod_ssi
 ##
 #include "conf.d/ssi.conf"


gstrauss:
Thanks.  PR will be picked up soon.

fbrosson:
Cool!

BTW, sorry for not having sent yet my improved version of mod_geoip.
(I did not find the time to clean/improve some parts of my code...)

glensc:
i'd love to see if the engine can decide itself which order modules need to be loaded. i.e geoip says i need to be before ssi. there are lots of systems (php, apache, ... etc) where this needs to be ordered and distributions have to resort to using `conf.d` style file fragments with `01-`, `10-`, ... prefixes.

@gstrauss something doable?

gstrauss:
@glensc , is this an issue that people trip over frequently with lighttpd modules?  I am not aware of such.  Yes, there is a suggested order for loading lighttpd modules for expected functionality, mostly due which hooks run when, and which modules run during which hooks.

However, unlike some other web servers, lighttpd modules tend to be standalone and not interact much with other modules.  This may provide slightly more limited functionality compared to some other web servers, but it also tends to reduce complexity.

Is there evidence that module ordering is a hard problem for people to solve in lighttpd when they load lots of lighttpd modules?  I am interested in more info if this affects more than a quorum of people, but otherwise there are more important priorities on my list of requested lighttpd enhancements.

fbrosson:
I don't know if there would be situations were two different load sequences would produce slightly different functionality (and yet still make sense), but having an automated dependency resolver would make the configuration less flexible, unless, of course, extra parameters were added to allow an unusual ordering.

glensc:
@gstrauss well, i'm not saying it's highest! urgent! priority! task that should be implemented or that it's most common problem with lighttpd setups.

i've done some initial ordering fixes in ~2006 and it has not needed much changes in that area afterwards. but i might not need the changes because i got lucky (geo < ssi in alphabet) and i do not use all modules in my setups.

$ ls /etc/lighttpd/conf.d/??_*.conf
- 10_mod_rewrite.conf
- 11_mod_redirect.conf
- 50_mod_access.conf
- 50_mod_alias.conf
- 50_mod_auth.conf
- 50_mod_cgi.conf
- 50_mod_cgi_php.conf
- 50_mod_cml.conf
- 50_mod_deflate.conf
- 50_mod_dirlisting.conf
- 50_mod_evasive.conf
- 50_mod_evhost.conf
- 50_mod_fastcgi.conf
- 50_mod_flv_streaming.conf
- 50_mod_geoip.conf
- 50_mod_indexfile.conf
- 50_mod_mysql_vhost.conf
- 50_mod_proxy.conf
- 50_mod_rrdtool.conf
- 50_mod_scgi.conf
- 50_mod_secdownload.conf
- 50_mod_setenv.conf
- 50_mod_simple_vhost.conf
- 50_mod_ssi.conf
- 50_mod_staticfile.conf
- 50_mod_status.conf
- 50_mod_uploadprogress.conf
- 50_mod_userdir.conf
- 50_mod_usertrack.conf
- 50_mod_webdav.conf
- 55_mod_expire.conf
- 55_mod_magnet.conf
- 60_mod_compress.conf
- 90_mod_accesslog.conf
- 91_mod_extforward.conf

but i dislike that part that such manual module load order must be done at all.

of possible solutions, perhaps add some internal priority for each hook that module has, some number in range -100 ... 100., so that such problems as:

> In modules.conf, mod_geoip needs to be loaded before mod_ssi, otherwise GeoIP vars won't be available to SSI pages.

can be solved with (default priority is `0`):
- mod_geoip: { .export_http_vars: { .priority: -50, handler: handler } }
- mod_ssi: { .export_http_vars: NULL }

in [my distribution](https://github.com/pld-linux/lighttpd) each lighttpd module has it's own conf.d config that does `server.modules += ( "mod_<modname>" )` and defines rest of module specific variables with few defaults resulting highly modular setup when each module can be enabled/disabled.

gstrauss:
As already mentioned, most modules do not interact.  mod_rewrite, mod_cml, and mod_magnet are the most likely to interact since they might restart requests.  Any dynamic handlers all run in the subrequest hook and there can be only one handler, so those reset state if they restart the request (mod_fastcgi and mod_cgi might).  The only reason mod_geoip and mod_ssi ordering matter is because mod_ssi currently is in the subrequest_start hook, but ought to move its handling into the subrequest hook, at which point there would be no potential ordering issue with mod_geoip and mod_ssi.  There are some open feature requests for mod_ssi, and mod_ssi is likely to move its handling to the subrequest handler in the future.

The reason I mentioned priorities when talking about this suggestion of yours is that there are multiple ways to do this, and multiple workarounds if issues occur, and your specific use of a directory of *.conf is one way to do it (among a number of options), and your prefixing of the files with numbers is a valid way to control module ordering when you have a directory of *.conf files which I presume are globbed to be loaded.  Another option is to list the modules in a specific order in the lighttpd.conf file, instead of globbing.

In summary, while this would be a "nice to have" for some people, there are so many valid and existing ways to do this that I hardly see why it worth discussing further, unless -- as I mentioned in my first post -- this is causing problems for people, and I was not previously aware.

fbrosson:
I feel a little bit of electricity, although there is imho no reason for it. I'm sure @glensc also knows an automated dependency solver would be a _"nice to have"_ ; he was just showing that, like most distros, the one he uses has everything that is needed to ensure the modules are loaded in a correct order.
I apologize for this whole message. You are both so much more experienced developers than me, that I hardly explain how did I dare to write this message!!! ;-)

gstrauss:
@fbrosson, I give more currency to the opinions of those who produce code behind their ideas, and less to those dictating work to others with little justification for the work besides "nice to have".  If someone spends some of their own time attempting to implement something, it lends some credence to the proposition that it is a problem worth solving.  Otherwise, I have an existing list of problems that are worth solving, and the "nice to have" without justification or priority is not on that list, primarily because of the missing justification and priority.  Sure, it would be nice to have.

fbrosson:
I **totally** agree with you that this is **_just_** a _"nice to have"_ and that, given the number of more important things, this "_nice to have_" will remain a "_nice to have_" forever, and I believe this is not a problem for anyone ;-)

PR #74, opened 2016-12-04
[WIP] Add I2P support

str4d:
This PR adds native I2P support via the [SAM API](https://geti2p.net/en/docs/api/samv3). It considerably simplifies the setup process for a server admin wanting to make their site accessible over I2P.

Still to do:
- Fix integration with `fdevents` subsystem.
- Add default-on enforcement of no `$HTTP["host"]` inside a `$SERVER["socket"] == "i2p:*"` to help prevent [hostname hacking](https://mascherari.press/onionscan-report-this-one-weird-trick-can-reveal-information-from-25-of-the-dark-web-2/).

To build:
- Copy `libsam3.c` and `libsam3.h` into `src/`
  - Currently hosted inside I2P
  - Perhaps these should just be included in the PR? (1073 and 254 lines respectively)
- `./configure --enable-i2p`

Example usage:
```
server.bind = "i2p:test_server"
server.port = 8080
server.document-root = "/path/to/pages/"
server.errorlog = "/path/to/error.log"
server.modules = ( "mod_accesslog" )
accesslog.filename = "/path/to/access.log"
accesslog.format  = "%{X-I2P-DestB64}i %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\""
i2p.sam.port = 7656
i2p.sam.keydir = "/path/to/keys"

$SERVER["socket"] == "i2p:second_server:8081" {
       i2p.sam.nickname = "Alternative server"
}
```

### DIFF
--- a/src/connections-glue.c
+++ b/src/connections-glue.c
@@ -227,47 +227,6 @@ int connection_handle_read(server *srv, connection *con) {
 	size_t mem_len = 0;
 	int toread;
 
-#ifdef HAVE_I2P
-	if (con->srv_socket->is_i2p && con->reading_i2p_dest) {
-		/* Read the first line, containing the remote Destination, before anything else happens */
-		char cur[2];
-		while (con->reading_i2p_dest && (1 == read(con->fd, cur, 1))) {
-			if (cur[0] == '\n') {
-				con->reading_i2p_dest = 0;
-			} else if (cur[0] == ' ') {
-				/* Not an I2P Destination -> not via SAM */
-				log_error_write(srv, __FILE__, __LINE__, "s", "Non-I2P request received on I2P socket");
-				connection_set_state(srv, con, CON_STATE_ERROR);
-				return -1;
-			} else {
-				buffer_append_string_len(con->i2p_dest, cur, 1);
-			}
-		}
-		if (con->reading_i2p_dest) {
-			return 0;
-#ifdef USE_OPENSSL
-		} else {
-			/* Calculate hash */
-			int raw_dest_len;
-			unsigned char *raw_dest = i2p_unbase64(con->i2p_dest->ptr, strlen(con->i2p_dest->ptr), &raw_dest_len);
-			unsigned char *hash = SHA256(raw_dest, raw_dest_len, 0);
-			free(raw_dest);
-			unsigned char *hash_b64 = i2p_base64(hash, strlen(hash), &raw_dest_len);
-			buffer_copy_string(con->i2p_dest_hash, hash_b64);
-			free(hash_b64);
-
-			/* Calculate B32 */
-			char b32_hash[56];
-			sam3Base32Encode(b32_hash, hash, strlen(hash));
-			char *eq_pos = strchrnul(b32_hash, '=');
-			eq_pos[0] = '\0';
-			buffer_copy_string(con->i2p_dest_b32, b32_hash);
-			buffer_append_string_len(con->i2p_dest_b32, CONST_STR_LEN(".b32.i2p"));
-#endif
-		}
-	}
-#endif
-
 	if (con->srv_socket->is_ssl) {
 		return connection_handle_read_ssl(srv, con);
 	}
--- a/src/connections-glue.c
+++ b/src/connections-glue.c
@@ -10,6 +10,7 @@
 #ifdef USE_OPENSSL
 # include <openssl/ssl.h>
 # include <openssl/err.h>
+# include <openssl/sha.h>
 #endif
 
 const char *connection_get_state(connection_state_t state) {
@@ -225,6 +226,47 @@ int connection_handle_read(server *srv, connection *con) {
 	char *mem = NULL;
 	size_t mem_len = 0;
 	int toread;
+
+#ifdef HAVE_I2P
+	if (con->srv_socket->is_i2p && con->reading_i2p_dest) {
+		/* Read the first line, containing the remote Destination, before anything else happens */
+		char cur[2];
+		while (con->reading_i2p_dest && (1 == read(con->fd, cur, 1))) {
+			if (cur[0] == '\n') {
+				con->reading_i2p_dest = 0;
+			} else if (cur[0] == ' ') {
+				/* Not an I2P Destination -> not via SAM */
+				log_error_write(srv, __FILE__, __LINE__, "s", "Non-I2P request received on I2P socket");
+				connection_set_state(srv, con, CON_STATE_ERROR);
+				return -1;
+			} else {
+				buffer_append_string_len(con->i2p_dest, cur, 1);
+			}
+		}
+		if (con->reading_i2p_dest) {
+			return 0;
+#ifdef USE_OPENSSL
+		} else {
+			/* Calculate hash */
+			int raw_dest_len;
+			unsigned char *raw_dest = i2p_unbase64(con->i2p_dest->ptr, strlen(con->i2p_dest->ptr), &raw_dest_len);
+			unsigned char *hash = SHA256(raw_dest, raw_dest_len, 0);
+			free(raw_dest);
+			unsigned char *hash_b64 = i2p_base64(hash, strlen(hash), &raw_dest_len);
+			buffer_copy_string(con->i2p_dest_hash, hash_b64);
+			free(hash_b64);
+
+			/* Calculate B32 */
+			char b32_hash[56];
+			sam3Base32Encode(b32_hash, hash, strlen(hash));
+			char *eq_pos = strchrnul(b32_hash, '=');
+			eq_pos[0] = '\0';
+			buffer_copy_string(con->i2p_dest_b32, b32_hash);
+			buffer_append_string_len(con->i2p_dest_b32, CONST_STR_LEN(".b32.i2p"));
+#endif
+		}
+	}
+#endif
 
 	if (con->srv_socket->is_ssl) {
 		return connection_handle_read_ssl(srv, con);
--- a/configure.ac
+++ b/configure.ac
@@ -847,6 +847,23 @@ if test x$ipv6 = xtrue; then
   fi
 fi
 
+dnl Check for I2P support
+
+AC_MSG_CHECKING(for I2P support)
+AC_ARG_ENABLE(i2p,
+ AC_HELP_STRING([--enable-i2p],[enable I2P support]),
+ [case "${enableval}" in
+   yes) i2p=true ;;
+    no) i2p=false ;;
+     *) AC_MSG_ERROR(bad value ${enableval} for --enable-i2p) ;;
+  esac],[i2p=false])
+AC_MSG_RESULT($enableval)
+
+if test x$i2p = xtrue; then
+  AC_DEFINE(HAVE_I2P,1,[Whether to enable I2P support])
+fi
+AM_CONDITIONAL(BUILD_WITH_I2P, test x$i2p = xtrue)
+
 # disable mmap by default; if a mmapped file gets truncated, the process gets a SIGBUS signal
 # on reading the truncated area which we can't handle (yet).
 # lighttpd may always use mmap with files it owns (created tmp files)
@@ -1018,6 +1035,13 @@ fi
 
 features="network-ipv6"
 if test "$ac_cv_ipv6_support" = yes; then
+	enable_feature="$enable_feature $features"
+else
+	disable_feature="$disable_feature $features"
+fi
+
+features="network-i2p"
+if test "x$i2p" = xtrue; then
 	enable_feature="$enable_feature $features"
 else
 	disable_feature="$disable_feature $features"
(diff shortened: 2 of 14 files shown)


str4d:
The integration is fully-functional as of 776d88151e2c589948b48f9c0fc02311ff1098e5, but only with I2P routers on the same machine (otherwise each listener configured needs to be manually port-forwarded from the I2P router to the webhost, which is no better in terms of usability that current methods). Using `STREAM ACCEPT` fixes this, but I don't understand the `fdevents` subsystem well enough to fix the segfault issues that result from 001a61210522c3cc0511b74e14d500809ab68106.

gstrauss:
Interesting.  Thanks.  This will take a bit of time to review.

First glance: lighttpd already has base64 encode and decode routines (although you're more than welcome to level criticism against the implementations).  Is there a reason you include another implementation (i2p-base64.h taken from https://github.com/superwills/NibbleAndAHalf)?  If one implementation is superior, it should replace the other.  lighttpd should not have duplicated base64 implementations.

str4d:
@gstrauss Honestly, I don't remember. I wrote the majority of this in July 2015 (up to a working `STREAM FORWARD`), on top of 1.4.35 - perhaps there wasn't a base64 implementation then? Or more likely, I just didn't spot it - I've been learning the lighttpd codebase as I go. The primary requirement is that the two non-alphanumeric characters can be swapped out for the URL-safe ones that I2P uses (which some libraries support via their API; the small one I picked doesn't).

src/connections-glue.c
```diff
@@ -10,6 +10,7 @@
 #ifdef USE_OPENSSL
 # include <openssl/ssl.h>
 # include <openssl/err.h>
+# include <openssl/sha.h>
```
str4d: This was supposed to be added to `src/connections.c` 😅 (although perhaps the corresponding code should be moved into `src/network_i2p.c` now that it exists?)

gstrauss:
**This should be a standalone module, without sprinkling itself across so many files;** overall this needs a lot of work and can be much better isolated.

Some hints: write your own network_server_handle_fdevent() and connection_accept(), call connection_accepted(), make i2p-specific modifications on the returned 'con', then call connection_state_machine().

Second hint: instead of extending $SERVER["socket"], all the i2p directives could be parsed out of a single list directive which contains an array of key => value, e.g. see mod_scgi.c scgi.server = ( ... ), and you could have the i2p listener address as one of the keys ("listen" => "...addr...").

For full integration, we may find that there needs to be a hook for listener initialization, overload conditions, and graceful shutdown, but we can consider those requirements after this is an isolated, standalone module.

str4d:
Thanks for the hints. I originally tried to write this as a standalone module (which I agree would have been much better), but at the time (July 2015) it appeared to me that the socket/network layer was the one part of the codebase that could not be hooked into, as it assumed that the only network transports were IPv4, IPv6 and Unix sockets. If that is no longer the case, excellent!

Where should I look for documentation of the fdevent subsystem (specifically, intended usage and failure conditions, to figure out why my current code results in a condition hard-coded to give a segfault)?

> all the i2p directives could be parsed out of a single list directive which contains an array of key => value, e.g. see mod_scgi.c scgi.server = ( ... ), and you could have the i2p listener address as one of the keys ("listen" => "...addr...").

I'm not sure that would work, because the following configuration would AFAICT be invalid:

```
# Site A, served inside I2P
server.docroot = ".../site-a/"
i2p.server = (
    "listen" => "site_a_dest"
)
# Site B, served inside I2P
# Not possible to have a second server defined here?
i2p.server = (
    "listen" => "site_b_dest"
    "server.docroot" => ".../site-b/"
)
# Site C, served both on clearnet and inside I2P
$HTTP["host"] == "site-c.com" {
    server.docroot = ".../site-c/"
    i2p.server = (
        "listen" => "site_c_i2p_dest"
    )
}
```

This is the kind of setup that I want to enable with an I2P module, and without being able to hook into `$SERVER["socket"]` for conditionally hosting sites, I don't see how it would work (since there doesn't appear to be support for modules to extend [conditional fields](https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_Configuration). With the current PR, I'd do the following:

```
# Site A, served inside I2P
$SERVER["socket"] == "i2p:site_a_dest" {
    server.docroot = ".../site-a/"
}
# Site B, served inside I2P
$SERVER["socket"] == "i2p:site_b_dest" {
    server.docroot = ".../site-b/"
}
# Site C, served both on clearnet and inside I2P
$HTTP["host"] == "site-c.com" {
    server.docroot = ".../site-c/"
}
$SERVER["socket"] == "i2p:site_c_dest" {
    # Not ideal to define parallel configs
    # (AFAICT a module would improve this by enabling nesting)
    server.docroot = ".../site-c/"
}
```

gstrauss:
I was thinking more along the lines of scgi.server config of nested lists
i2p.server = (
  ("site-a" => ("listen" => "site_a_dest", x => y, a => b, ... ) ),
  ("site-b" => ("listen" => "site_b_dest", x => y, a => b, ... ) ),
)
However, if you need to set other lighttpd server directives for other modules, not specific to i2p, then yes, there would need to be a way to configure them inside conditions.  I am not familiar with i2p, but since you were extending $SERVER["socket"], I surmise it is based on the address.  In that case, there is another feature request in the issue queue requesting the ability to make conditions based on local IP address (the listen address), and I might consider adding that.  (There is already a means to do so for the remote IP.)  In the meantime, I would suggest focusing on getting things working well with a single i2p server before worrying about how to configure conditions for multiple servers.

Regarding base64, are you unable to use one of the existing charsets that you're looking to add another?

> Where should I look for documentation of the fdevent subsystem (specifically, intended usage and 
failure conditions, to figure out why my current code results in a condition hard-coded to give a segfault)?

The fdevent system is a thin wrapper around socket events.  Since you have so far neglected to given any details besides "it aborts", it is hard to say why your code causes aborts.  Ultimately, you need to play nicely with the fdevent subsystem, using the fdevent interface to change interest in socket events on the fd, and to register and unregister the fd.  Breaking the fdevent accounting of the fd is likely what causes the abort.  More recent versions of lighttpd have moved the fdevent code in the direction of better encapsulation, but there is more work to do.  If you haven't looked at the code since 2015, you have quite a bit of catching up to do. ;)  Look at some of the simpler dynamic handlers, such as mod_proxy for managing a socket to a backend.

str4d:
Disclaimer: I'm an I2P developer 😃

> I was thinking more along the lines of scgi.server config of nested lists
> However, if you need to set other lighttpd server directives for other modules, not specific to i2p, then yes, there would need to be a way to configure them inside conditions. I am not familiar with i2p, but since you were extending $SERVER["socket"], I surmise it is based on the address.

Just to ensure we are on the same page, this is what I imagine as an SCGI setup:

```
IPv4 ----                 ---- SCGI
         \___ lighttpd __/
IPv6 ----/               \---- SCGI
```

whereas this is what I'm trying to achieve with this PR:

```
IPv4 ----
I2P -----\___ lighttpd
IPv6 ----/
```

I2P is a network overlay. Essentially, you can treat an I2P Destination as *roughly* equivalent to an IP address. Hence, if native support is added, the intention is that the resulting config syntax could transparently handle I2P Destinations (represented in the config by references to private key files) alongside IPv4, IPv6 and Unix sockets.

For context, the way that I use lighttpd with I2P currently is:

- Set up a server "tunnel" in I2PTunnel that forwards connections from an I2P Destination (inside I2P) to a configured `host:port` (outside I2P).
- Set up lighttpd to listen on `host:port`, possibly using vhosts (since I2PTunnel can be configured to set `Host:` appropriately).

This has several drawbacks:

- It requires either running the I2P router and lighttpd on the same machine, or making each lighttpd listener accessible to the router (via port-forwarding or specific network configuration).
- It is not a particularly seamless experience for lighttpd server admins (my ulterior motive with this work is to make it easier for server admins to use I2P by leveraging the tools they are familiar with).
- Most importantly, lighttpd has no "knowledge" of the fact that a connection is arriving from I2P, meaning that it is much easier for server admins to make mistakes that can compromise the privacy benefits.

> In that case, there is another feature request in the issue queue requesting the ability to make conditions based on local IP address (the listen address), and I might consider adding that. (There is already a means to do so for the remote IP.) In the meantime, I would suggest focusing on getting things working well with a single i2p server before worrying about how to configure conditions for multiple servers.

If that can be hooked by a module, it should do the trick.

> Regarding base64, are you unable to use one of the existing charsets that you're looking to add another?

I2P has an existing charset that it has used since at least 2005, possibly earlier. This is how Destinations are text-encoded everywhere, including when sent over the SAM API. So to convert them into the easier-to-handle-in-logs B32 format, it is necessary to be able to decode from the I2P charset.

> The fdevent system is a thin wrapper around socket events. Since you have so far neglected to given any details besides "it aborts", it is hard to say why your code causes aborts.

Well, I didn't want to waste your time debugging it, but if you insist 😉

The way that `STREAM ACCEPT` works is that a connection is opened to the SAM API telling it to listen, and then when a client connects, the SAM API forwards their I2P data back through the connection. Based on how the regular IPv* sockets are hooked into fdevent, it looked like I needed to [register the listening fds for nework server events](https://github.com/lighttpd/lighttpd1.4/pull/74/commits/001a61210522c3cc0511b74e14d500809ab68106#diff-454300ae5483905fb14fa49f2cc1c45aR250), and then when data is available (a new connection), [unregister the fd from fdevent](https://github.com/lighttpd/lighttpd1.4/pull/74/commits/001a61210522c3cc0511b74e14d500809ab68106#diff-454300ae5483905fb14fa49f2cc1c45aR279) so that it will later be re-registered inside `connection_accepted` for connection events. But the event deletion call triggers [this SEGFAULT](https://github.com/lighttpd/lighttpd1.4/blob/master/src/fdevent_linux_sysepoll.c#L40) with "Bad file descriptor", despite the file descriptor in question being identical to when it was originally created.

(If I comment out the unregistering calls, then when a client connects I get the same segfault but with a different errno, triggered from [this call in the state machine](https://github.com/lighttpd/lighttpd1.4/blob/master/src/connections.c#L1424). I figure that's a side-effect of misusing the fdevent subsystem, hence me trying to learn more about it.)

> Ultimately, you need to play nicely with the fdevent subsystem, using the fdevent interface to change interest in socket events on the fd, and to register and unregister the fd. Breaking the fdevent accounting of the fd is likely what causes the abort. More recent versions of lighttpd have moved the fdevent code in the direction of better encapsulation, but there is more work to do. If you haven't looked at the code since 2015, you have quite a bit of catching up to do. ;) Look at some of the simpler dynamic handlers, such as mod_proxy for managing a socket to a backend.

Will do, although as I mentioned earlier, I2P is not a backend - if only it were that easy! 😄

gstrauss:
Your patches add libsam3.[ch] to the build configuration, but they are not part of your commits.  Are they supposed to be installed separately?  Your additions to the build configuration do not match the existing build flags.  --enable-i2p should probably be --with-i2p and should not require an argument to enable, though it should be "no" by default in configure.ac.

The code should build warnings-clean, and so if there are warnings when you compile, you might look there for why you might be crashing.

gstrauss:
> First pass at I2P support
> Requires http://git.repo.i2p/w/libsam3.git - copy src/libsam3/* into src/
> commit 5d776f4e8a587de03a9a132d38f109e5f78db1c4 @str4d str4d committed on Jul 4, 2015

That URL is not valid.

https://geti2p.net/en/docs/api/samv3 links to http://git.repo.i2p.xyz/w/libsam3.git but that site seems to be misconfigured.
```
$ git clone http://git.repo.i2p.xyz/w/libsam3.git
Cloning into 'libsam3'...
fatal: http://git.repo.i2p.xyz/w/libsam3.git/info/refs not valid: is this a git repository?
```

http://git.repo.i2p.xyz/w
500 - Internal Server Error
Can't locate object method "startform" via package "CGI" at /srv/eep/git.repo.i2p.xyz/cgi-bin/gitweb.cgi line 4811.

gstrauss:
(If I have to set up i2p and run i2p binaries before I can see the source i2p code, that's a non-starter.)

gstrauss:
After disussion on IRC, withdrawing this pull request.  @str4d, please create a PR in your own lighttpd fork and send an email to me.  Discussion can continue there.

The current code in this PR has numerous deficiencies, but we can't even discuss them until after the code is more complete.  Please mirror the libsam3 code to a repo on github somewhere, or other location that is not solely accessible via the i2p network.

The libsam3 code is missing from this PR.  From older (unofficial?) versions, the libsam3 code appears to be licenced under LGPL or GPL3 or ???.  There is no chance that it will be accepted into lighttpd as part of this pull request due to that license being more restrictive than the lighttpd BSD license.  Were libsam3 available as a standalone library that users could install separately, and the code in this PR licensed in a BSD-compatible way, there is a chance that some of it could be accepted into the lighttpd repository.

str4d:
FYI the libsam3 I was working with is licensed under the WTFPL (by the original dev), hence why the branch you found could have been relicensed under the LGPL. But it can be easily re-licensed to a more standard permissive license.

PR #77, opened 2017-03-15
[core] allow overriding previously set configuration values

pprindeville:
x-ref:
  "allow overriding configuration values"
  https://redmine.lighttpd.net/issues/2799

### DIFF
--- a/src/configparser.y
+++ b/src/configparser.y
@@ -190,15 +190,8 @@ varline ::= key(A) ASSIGN expression(B). {
           ctx->current->context_ndx,
           ctx->current->key->ptr, A->ptr);
       ctx->ok = 0;
-    } else if (NULL == array_get_element(ctx->current->value, B->key->ptr)) {
-      array_insert_unique(ctx->current->value, B);
-      B = NULL;
     } else {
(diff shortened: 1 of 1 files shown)


gstrauss:
Thanks for putting together a patch.  Unfortunately, it can not be accepted.  While it might work for you the way that you would like it to work, it may break existing usage where breakage is expected if multiple (independent) /etc/lighttpd/conf.d/*.conf files incorrectly use the same directive in the same scope.

As mentioned in https://redmine.lighttpd.net/issues/2799, a simple solution is to maintain your lighttpd.conf and includes separately from the distro-installed configs, and to modify the startup scripts (or create a new service with new startup scripts) to start `lighttpd -f /path/to/your/custom/lighttpd.conf`

pprindeville:
> While it might work for you the way that you would like it to work, it may break existing usage where breakage is expected if multiple (independent) /etc/lighttpd/conf.d/*.conf files incorrectly use the same directive in the same scope.

Okay, here's a compromise: introduce another knob that controls whether previous values can be changed or not.  I can redo the PR and incorporate that.

Then there's zero chance of anyone accidentally breaking configs, since they'd have to deliberately enable this behavior.

PR #78, opened 2017-03-21
    [core] allow overriding previously set configuration values

pprindeville:
x-ref:
      "allow overriding configuration values"
      https://redmine.lighttpd.net/issues/2799

### DIFF
--- a/src/configfile.c
+++ b/src/configfile.c
@@ -967,6 +967,14 @@ static int config_tokenizer(server *srv, tokenizer_t *t, int *token_id, buffer *
 			}
 			break;
 
+		case ':':
+			if (t->input[t->offset+1] == '=') {
+				t->offset += 2;
+				tid = TK_FORCE_ASSIGN;
+				buffer_copy_string_len(token, CONST_STR_LEN(":="));
+			}
+			break;
+
 		case '{':
 			t->offset++;
 
--- a/src/configparser.y
+++ b/src/configparser.y
@@ -206,6 +206,23 @@ varline ::= key(A) ASSIGN expression(B). {
   A = NULL;
 }
 
+varline ::= key(A) FORCE_ASSIGN expression(B). {
+  if (ctx->ok) {
+    if (strncmp(A->ptr, "env.", sizeof("env.") - 1) == 0) {
+      fprintf(stderr, "Setting env variable is not supported in conditional %d %s: %s\n",
+	ctx->current->context_ndx,
+	ctx->current->key->ptr, A->ptr);
+      ctx->ok = 0;
+    } else {
+      buffer_copy_buffer(B->key, A);
+      array_replace(ctx->current->value, B);
+      B = NULL;
+    }
+  }
+  buffer_free(A);
+  A = NULL;
+}
+
 varline ::= key(A) APPEND expression(B). {
   if (ctx->ok) {
     array *vars = ctx->current->value;


gstrauss:
Thank you for putting together and proposing a patch to implement your desired feature.

I have followed the conversation on IRC and while I am not an expert in UX, I do have some consternation about adding a new keyword to the parser.  I am loathe to extend the *parsing syntax* in new ways.  A new token (e.g. ":=" as suggested on IRC) is more amenable to me.

Untested:
```
--- a/src/configfile.c
+++ b/src/configfile.c
@@ -967,6 +967,14 @@ static int config_tokenizer(server *srv, tokenizer_t *t, int *token_id, buffer *
                        }
                        break;
 
+               case ':':
+                       if (t->input[t->offset + 1] == '=') {
+                               t->offset += 2;
+                               buffer_copy_string_len(token, CONST_STR_LEN(":="));
+                               tid = TK_FORCE_ASSIGN;
+                       }
+                       break;
+
                case '{':
                        t->offset++;
 
diff --git a/src/configparser.y b/src/configparser.y
index 9def054..5891134 100644
--- a/src/configparser.y
+++ b/src/configparser.y
@@ -206,6 +206,23 @@ varline ::= key(A) ASSIGN expression(B). {
   A = NULL;
 }
 
+varline ::= key(A) FORCE_ASSIGN expression(B). {
+  if (ctx->ok) {
+    if (strncmp(A->ptr, "env.", sizeof("env.") - 1) == 0) {
+      fprintf(stderr, "Setting env variable is not supported in conditional %d %s: %s\n",
+          ctx->current->context_ndx,
+          ctx->current->key->ptr, A->ptr);
+      ctx->ok = 0;
+    } else {
+      buffer_copy_buffer(B->key, A);
+      array_replace(ctx->current->value, B);
+      B = NULL;
+    }
+  }
+  buffer_free(A);
+  A = NULL;
+}
+
 varline ::= key(A) APPEND expression(B). {
   if (ctx->ok) {
     array *vars = ctx->current->value;
```

Putting your patch in context: in a prior request at https://redmine.lighttpd.net/issues/2799, you expressed a desire to attempt to overlay config file snippets where some parts are maintained by a distro and some are maintained by you.  This complexity *is not* recommended.  While it is definitely fair to say that there are more than a few historical limitations in the lighttpd 1.4.x config parsing, it is possible to have a very short and simple configuration to have lighttpd perform as a basic web server.  This simplicity for basic functionality is a good thing.

The recommended solution to customize your lighttpd.conf is to maintain your lighttpd configuration separately from that of the sample configuration provided by your distro, which is easy to do and merely involves changing your lighttpd startup script to point to your config file entry point instead of to the lighttpd.conf provided by the distro.  Your concern about wanting to overlay (instead of maintaining a separate custom config) in order to magically inheriting distro lighttpd.conf changes is specious, since lighttpd.conf config file changes to a base lighttpd config file provided by most distros (at least those that I follow) is exceedingly rare.  As already mentioned, the basic lighttpd config is fairly simplistic, including the one provided by openwrt.

If this patch is to be accepted into lighttpd (in one form or another), then it will be accepted as an experimental feature until more and wider feedback is received on how well-behaved this feature is, and whether or not it is confusing or causes unforseen problems.  As an experimental feature, it may be removed should there be conflicts that are determined to be not-backwards-compatible with existing lighttpd behavior, or large headaches to fix or maintain.  "Bugs" will not be accepted for experimental features, but feature requests to fix or modify the behavior will be accepted for review.

(It is unfortunate that you have given me reason to spell this all out so explicitly to you given the assumptions that you brought into the conversation without first attempting to understand the language and terminology used to communicate effectively in *this* open source project.)

gstrauss:
One more thing: I will not support adding this feature as an independent patch to openwrt, separate from upstream lighttpd.  While you may be of the opinion that this feature is a dire need for you -- and that is fine for you to judge your own needs and your own systems -- I am unconvinced of the importance that this be maintained as a standalone patch.  That said, I am open to incorporating this into upstream lighttpd as an experimental feature, as I have posted above.

pprindeville:
I'm happy with either suggestion, either the `override` keyword (_a la_ Make) or with a separate assignment operator (oddly, also like Make).

pprindeville:
Okay, updated.

PR #84, opened 2017-08-16
[mod_openssl] Allow to selectively disable TLS 1.0, 1.1 and 1.2 versions

lima-j:
Up until now, lighttpd allowed to disable SSLv2 and SSLv3 through the ssl.use-sslv2 and ssl.use-sslv2 lighttpd.conf options, however the same was not offered to selectively disable TLS versions.

On the lighttpd forum, only enabling the TLS 1.2 ciphers is suggested as a mitigation to disable TLS 1.0 (https://redmine.lighttpd.net/boards/2/topics/5797), however it disables TLS 1.1 as well.

This patch adds the option to selectively disable the TLS versions (using OpenSSL's SSL_OP_NO_TLSv1, SSL_OP_NO_TLSv1_1, SSL_OP_NO_TLSv1_2 parameters), without having to change the cipher suite configuration, by adding the ssl.use-tlsv10, ssl.use-tlsv11 and ssl.use-tlsv12 lighttpd.conf options.

### DIFF
--- a/doc/config/lighttpd.conf
+++ b/doc/config/lighttpd.conf
@@ -485,6 +485,17 @@ server.upload-dirs = ( "/var/tmp" )
 #ssl.ca-crl-file = ""
 
 ##
+## it is also possible to selectively enable/disable SSL/TLS versions, by setting the following parameters accordingly.
+## TLS v1.1 and TLS v1.2 are enabled by default.
+##
+##
+#ssl.use-sslv2 = "disable"
+#ssl.use-sslv3 = "disable"
+#ssl.use-tlsv10 = "disable"
+#ssl.use-tlsv11 = "enable"
+#ssl.use-tlsv12 = "enable"
+
+##
 #######################################################################
 
 #######################################################################
--- a/src/mod_openssl.c
+++ b/src/mod_openssl.c
@@ -55,6 +55,9 @@ typedef struct {
     unsigned short ssl_empty_fragments; /* whether to not set SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS */
     unsigned short ssl_use_sslv2;
     unsigned short ssl_use_sslv3;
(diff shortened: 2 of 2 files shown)


gstrauss:
Thanks for the patch.  However, I'm looking to combine these directives into a single directive rather than having an ever-increasing number of directives to manage "allowed SSL/TLS versions".  I might have some time to prototype something in Sept, but you're more than welcome to have a go at it.

lima-j:
Hi Glenn,

Thanks for your response. I assume that you won't accept this patch then, is that correct?

Could you elaborate a bit on what you envision for that single directive thingy? I can try to give it a look in the meantime, if it is of any help to you.

lima-j:
Hi Glenn,

I haven't got a response from you to my previous message.

I may have some time to contribute a definitive fix for this issue, if you're not going to accept the one I'm pulling here, but would like to hear more from you on what you expect it to be.

stbuehler:
Not sure what @gstrauss has in mind, but maybe something based on `SSL_CONF_*` (see https://wiki.openssl.org/index.php/Manual:SSL_CONF_cmd(3)).

gstrauss:
@lima-j sorry I have not had a chance to get to this yet.  Yes, @stbuehler is on the mark.  Rather than a growing list of directives to address the multitude of options, I'd like to provide some interface to something like https://wiki.openssl.org/index.php/Manual:SSL_CONF_cmd(3)

gstrauss:
@lima-j  please have a look at https://github.com/gstrauss/lighttpd1.4/commit/c09acbeb8a030942d9825b3d0dd01c84e0a0b919 for new, experimental `ssl.openssl.ssl-conf-cmd` directive

lima-j:
@gstrauss it looks good, thanks for letting me know!!

PR #85, opened 2017-10-16
[core] Add back REQUEST_SCHEME (for use in e.g. mod_ssi)

fbrosson:
~~Also make it available to mod_accesslog as %S.~~

### DIFF
--- a/src/http-header-glue.c
+++ b/src/http-header-glue.c
@@ -1452,6 +1452,9 @@ int http_cgi_headers (server *srv, connection *con, http_cgi_opts *opts, http_cg
     rc |= cb(vdata, CONST_STR_LEN("GATEWAY_INTERFACE"),
                     CONST_STR_LEN("CGI/1.1"));
 
+    rc |= cb(vdata, CONST_STR_LEN("REQUEST_SCHEME"),
+                    CONST_BUF_LEN(con->uri.scheme));
+
     if (buffer_is_equal_caseless_string(con->uri.scheme,
                                         CONST_STR_LEN("https"))) {
         rc |= cb(vdata, CONST_STR_LEN("HTTPS"), CONST_STR_LEN("on"));


fbrosson:
Hi @stbuehler and @gstrauss !
`REQUEST_SCHEME` was added in a579e7ffc0d524886cf5c65c97284c2b2c3d958a but was lost in 7fa5bfc9381764d6c35ac29674bd830e8e5fee33.
This means `REQUEST_SCHEME` is in 1.4.**45** but won't be in 1.4.**46** _unless_ the tiny change in `src/http-header-glue.c` gets merged.

The changes in `src/mod_accesslog.c` aren't very important, unless one wishes to log the **`con->uri.scheme`** (i.e. `REQUEST_SCHEME`) with mod_accesslog.
BTW, I tried to use **`%{REQUEST_SCHEME}e`** in **`accesslog.format`** but it did not yeld anything. So I decided to make the changes in `src/mod_accesslog.c` to be able to use **`%S`** (instead of `%{REQUEST_SCHEME}e`) but I don't know if this is desirable.
Should I drop the changes in `src/mod_accesslog.c` and only keep the tiny change in `src/http-header-glue.c` ?
Thanks!

More or less related refs:
https://redmine.lighttpd.net/issues/2640
https://redmine.lighttpd.net/issues/2721

gstrauss:
Yes, REQUEST_SCHEME is provided with Apache and nginx for CGI, so there is some merit to that part of patch.  Thanks for catching that it was previously only for mod_ssi and was accidentally lost in the merge of CGI variables for all CGI-like backends.

On the other hand, %S means something different in http://httpd.apache.org/docs/2.4/mod/mod_log_config.html and so this patch is unlikely to be accepted as-is.  Apache does not have a specific log directive for scheme.   Given both these, yes, please drop this part of the patch.

gstrauss:
Please move the code in http-header-glue.c to be right above where HTTPS variable is conditionally set.

fbrosson:
> Please move the code in http-header-glue.c to be right above where HTTPS variable is conditionally set.

done

fbrosson:
> Given both these, yes, please drop this part of the patch

ok, done.
(I had missed your first message!!!)

src/http-header-glue.c
```diff
@@ -1452,6 +1452,9 @@ int http_cgi_headers (server *srv, connection *con, http_cgi_opts *opts, http_cg
     rc |= cb(vdata, CONST_STR_LEN("GATEWAY_INTERFACE"),
                     CONST_STR_LEN("CGI/1.1"));
 
+    rc |= cb(vdata, CONST_STR_LEN("REQUEST_SCHEME"),
+                    CONST_BUF_LEN(con->uri.scheme));
+
```
fbrosson: @gstrauss, when you wrote _"Please move the code in http-header-glue.c to be right above where HTTPS variable is conditionally set"_ did you mean **_without_** inserting a blank line here?
(If so, I'll remove that blank line, of course.)

gstrauss:
Merged.  Thanks!

fbrosson:
Cool, thanks!

PR #88, opened 2017-12-30
Disable RC4 explicitly as per RFC7465

codarrenvelvindron:
As per RFC7465, added RC4 to the default disabled weak ciphers.
https://tools.ietf.org/html/rfc7465

### DIFF
--- a/doc/config/lighttpd.conf
+++ b/doc/config/lighttpd.conf
@@ -442,6 +442,8 @@ server.upload-dirs = ( "/var/tmp" )
 ##     # https://www.ssllabs.com/projects/best-practices/index.html)
 ##     # - BEAST is considered mitigaed on client side now, and new weaknesses have been found in RC4,
 ##     #   so it is strongly advised to disable RC4 ciphers (HIGH doesn't include RC4)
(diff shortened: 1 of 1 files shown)


stbuehler:
I'd say if `HIGH` includes `RC4` in your openssl version you are screwed anyway. My version doesn't even include `3DES` anymore...

Imho the point of that recommended cipher string is that it is short and understandable. You'll find many other recommendations, e.g. the one from "better crypto" in https://github.com/BetterCrypto/Applied-Crypto-Hardening/blob/master/src/configuration/Webservers/lighttpd/10-ssl.conf:

    ssl.cipher-list = "EDH+CAMELLIA:EDH+aRSA:EECDH+aRSA+AESGCM:EECDH+aRSA+SHA256:EECDH:+CAMELLIA128:+AES128:+SSLv3:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!DSS:!RC4:!SEED:!IDEA:!ECDSA:kEDH:CAMELLIA128-SHA:AES128-SHA"

I probably disagree with some of their choices, but I can't even read that thing - I can only check what list of ciphers it ends up on my system.

PR #90, opened 2018-05-25
[mod_accesslog] IP address privacy

onnokort:
To respect a user's privacy and make him or her pseudonymous, extra
functionality is hereby added to mod_accesslog.

When enabled using the "accesslog.ip-randomize-interval" configuration
variable, it will randomize a user's IP by concatenating it with a
regularly changing random number
(the randomize interval), hash it (using SHA256), and print 64 bits of
the resulting hash into the log file as a hex string.

This should allow for log file and click stream analysis (continuity of
the resulting hash fingerprint) while respecting the privacy of visitors.

This functionality is only enabled when openssl support is compiled into
lighttpd.

### DIFF
--- a/doc/config/conf.d/access_log.conf
+++ b/doc/config/conf.d/access_log.conf
@@ -23,5 +23,24 @@ accesslog.filename          = log_root + "/access.log"
 ##
 #accesslog.use-syslog       = "enable"
 
+
+##
+## Specify whether and how often the request origin IP address should be
+## randomized in the log file
+##
+## If enabled, the IP address as a text string is concatenated with a
+## 128-bit random salt and SHA256-hashed, with 64 bits of the resulting hash
+## put into the log file instead of the clear text IP.
+##
+## Please specify the re-salting interval here. The salt is not
+## stored anywhere but in lighttpd's process RAM.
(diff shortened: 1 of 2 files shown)


glensc:
can you include example of the log output?

otherwise i'd say this functionality is not webserver responsibility. use logging to pipe and create program that obfuscates the field when reading from the pipe.

onnokort:
Example log line:

```284e34a5ce7a6e51 localhost:8000 - [25/May/2018:18:10:23 +0200] "GET /test/ HTTP/1.1" 404 345 "-" "Mozilla/5.0 ..."```

Fair point about using a pipe to an external process. I thought this change is small enough and makes it easier, but I can also use an external process, of course. In that case, just close the PR.

gstrauss:
> I thought this change is small enough
This change adds three (3!) directives and a dependency on openssl.  It is not as small as you think.

Using piped loggers is the recommended mechanism for those seeking to rewrite logs for GDPR compliance, as pipe logging for the lighttpd error log is also a concern.
https://redmine.lighttpd.net/boards/2/topics/8097

glensc:
🤣 for GDPR ;)

PR #92, opened 2018-09-21
Adds support for building Lighttpd with wolfSSL

dgarske:
Adds support for building Lighttpd with wolfSSL as a TLS and Crypto provider. Adds a new ./configure `--with-wolfssl` option.


doc/wolfssl/README
```diff
@@ -0,0 +1,127 @@
...
+
+Accept the complaints about unrecognized certs to see the page without server.
+Page should show same congratulatory text as above.
+
+4.0 Debugging
+
+To get debug messages from wolfSSL code you must configure with the `--enable-debug` option:
+    ./configure --enable-lighty --enable-debug
+
+To turn on the wolfSSL debugging the following must be added to lighttpd:
```
gstrauss: This belongs in mod_openssl.c in the mod_openssl_init() function, which is called when the module is loaded.  It does not belong in server.c

doc/wolfssl/README
```diff
@@ -0,0 +1,127 @@
...
+    wolfSSL_Debugging_ON();
+
+After you have done this, wolfSSL debug messages will be printed to stderr.
+
+The following line in the lighttpd.conf will specify the lighttpd log file:
+    server.errorlog = "./server_err.log"
+
+Prior to the log being opened, log messages will be directed to stderr.
+
+To use lighttpd logging the following function is called with a lighttpd
```
gstrauss: Why is this mentioned here?
"arguement" is spelled "argument"
While I do not particularly like the historical lighttpd log_error_write() interface, your example is not minimal.  The format string "s" indicates as single string following, so ... "s", "desired message"), not ... "ss", "", "desired message")

src/server.c
```diff
@@ -2048,6 +2048,10 @@ int main (int argc, char **argv) {
     }
   #endif
 
+  #ifdef DEBUG_WOLFSSL
```
gstrauss: This belongs in mod_openssl.c in the mod_openssl_init() function, which is called when the module is loaded.  The call does not belong in server.c

doc/wolfssl/README
```diff
@@ -0,0 +1,127 @@
...
+server *srv arguement:
+    log_error_write(srv, __FILE__, __LINE__, "ss", "", "desired message");
+
+
+4.1 PK Callbacks
```
gstrauss: Please name this specific to wolfSSL, e.g. --enable-wolfssl-pkcallbacks

configure.ac
```diff
@@ -565,6 +565,59 @@ fi
 AM_CONDITIONAL([BUILD_WITH_KRB5], [test "$WITH_KRB5" != no])
 AC_SUBST([KRB5_LIB])
 
+
+
```
gstrauss: Why are these here?  They are not bad, but it is not immediately obvious to me why this was added here.

src/mod_openssl.c
```diff
@@ -546,6 +551,163 @@ network_openssl_ssl_conf_cmd (server *srv, plugin_config *s)
   #endif
 }
 
+#if defined(HAVE_WOLFSSL_SSL_H) && defined(HAVE_PK_CALLBACKS)
```
gstrauss: Please name this specific to wolfSSL, e.g. HAVE_WOLFSSL_PKCALLBACKS
(How does --enable-pkcallbacks turn into HAVE_WOLFSSL_PK_CALLBACKS?)

doc/wolfssl/server.pem
```diff
@@ -0,0 +1,81 @@
+-----BEGIN RSA PRIVATE KEY-----
```
gstrauss: All over this PR, please do not include private keys.  People may copy configs and copy keys, even if these were meant only as an example.

Please instead provide instructions how to generate the certificate files.

src/data_config.c
```diff
@@ -210,6 +210,9 @@ int data_config_pcre_exec(data_config *dc, cond_cache_t *cache, buffer *b) {
         cache->comp_value = b; /* holds pointer to b (!) for pattern subst */
     return cache->patterncount;
 #else
+    (void)dc;
```
gstrauss: Compiling without warnings when PCRE is disabled (e.g. "./configure --without-pcre") should be a separate commit.  Multiple commits are permitted to be part of a single Pull Requests.  In any case, I will fix these separately.

src/first.h
```diff
@@ -55,4 +55,8 @@
 
 #define UNUSED(x) ( (void)(x) )
 
+#ifdef HAVE_WOLFSSL_SSL_H
```
gstrauss: This does not belong here.  There is only compiler and C library setup here.  What was your reasoning for adding something so wolfSSL specific here, to be included in every file?

If the other modules which check for "#if defined HAVE_LIBSSL && defined HAVE_OPENSSL_SSL_H" should be isolated into a new lighttpd header, such as sys-crypto.h, then let's create such an abstraction.

src/mod_authn_file.c
```diff
@@ -600,7 +600,15 @@ static void apr_sha_encode(const char *pw, char *result, size_t nbytes) {
     unsigned char digest[20];
     size_t base64_written;
 
+#ifdef HAVE_WOLFSSL_SSL_H
```
gstrauss: see above suggestion for sys-crypto.h

If wolfSSL does not provide SHA1(), then we can create a macro such as
```
#define SHA1(str,len,digest) do { SHA_CTX sha1; SHA1_Init(&sha1); SHA1_Update(&sha1, (const unsigned char *)(str), (len)); SHA1_Final((digest), &sha1); }
```
To avoid potential conflict with user-variable named 'sha1', I would use something a bit more unique, or make this a static inline function.  lighttpd is built to support (most of) C99.

src/mod_authn_file.c
```diff
@@ -646,6 +654,7 @@ static handler_t mod_authn_file_htpasswd_basic(server *srv, connection *con, voi
             #endif
            #endif
            #ifdef USE_OPENSSL_CRYPTO /* (for MD4_*() (e.g. MD4_Update())) */
```
gstrauss: Please add a comment that says that wolfSSL defines this if MD4 is not provided, e.g.
```
#ifndef NO_MD4 /*(defined if support for MD4 is not built into wolfSSL)*/
```

src/mod_dirlisting.c
```diff
@@ -87,6 +87,7 @@ static excludes_buffer *excludes_buffer_init(void) {
 	return exb;
 }
```
gstrauss: Incorrect (or at least misleading).  This looks to be part of warnings generated if lighttpd is built ./configure --without-pcre and should be addressed separately.

src/mod_secdownload.c
```diff
@@ -18,7 +18,15 @@
 #ifdef USE_OPENSSL_CRYPTO
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
-#endif
```
gstrauss: I have run into the CONST_STR_LEN() and CONST_BUF_LEN() issues with other inline macros, such as strncasecmp() and would rather can the affected two lines of code to replace CONST_BUF_LEN(), but without reaching into the buffer innards directly.  In other words, use buffer_string_length(), which is an inline function and the extra few instructions will be dwarfed by the HMAC calculations.

autogen.sh
```diff
@@ -10,5 +10,7 @@ fi
 
 # old autoreconf/aclocal versions fail hard if m4 doesn't exist
 mkdir -p m4
+# some versions fail if it doesn't exist
+touch ./config.rpath
```
gstrauss: Is this the right thing to do?  Would you provide more details?  (autoconf can be quite opaque and obtuse.  lighttpd does include some m4 scripts to address some shortcomings or limitations in some distributions.  Is there a better way to get past the error you are seeing?  When are you seeing an error?  What versions/platforms?)

Related, lighttpd supports multiple build systems, so if we add --with-wolfssl to configure.ac, we should also update the CMake, SCons, and meson build scripts, too.

gstrauss:
```
#if defined HAVE_LIBSSL && defined HAVE_OPENSSL_SSL_H
#define USE_OPENSSL_CRYPTO
#endif
```
Related to my comment about creating a sys-crypto.h, to enable use of wolfSSL, we may need to move the above logic from a few files into sys-crypto.h, too.

src/mod_authn_file.c
gstrauss: Actually, if wolfSSL does not provide SHA1() then it does not need to be an inline function.  It could simply be provided in algo_sha1.[ch].

src/mod_openssl.c
```diff
@@ -546,6 +551,163 @@ network_openssl_ssl_conf_cmd (server *srv, plugin_config *s)
...
+
+static int tls_create_key_cb(WOLFSSL* ssl, ecc_key* key, word32 keySz,
+    int ecc_curve, void* ctx)
+{
+    int     ret;
+    server* srv = (server*)ctx;
+    WC_RNG  rng;
+
+    (void)ssl;
```
gstrauss: Please use the UNUSED() macro for unused variables instead of explicitly casting to (void). In the future, the UNUSED() macro might include __attribute__((unused)) for compilers which support that attribute.

TimParrish:
Thank you for the feedback, we are working to address the comments and will push to this branch when they are ready.

doc/wolfssl/README
```diff
@@ -115,8 +118,8 @@ The following line in the lighttpd.conf will specify the lighttpd log file:
 Prior to the log being opened, log messages will be directed to stderr.
 
 To use lighttpd logging the following function is called with a lighttpd
-server *srv arguement:
-    log_error_write(srv, __FILE__, __LINE__, "ss", "", "desired message");
+server *srv argument:
+    log_error_write(srv, __FILE__, __LINE__, "s", "", "desired message");
```
gstrauss: The new code is incorrect and the last argument is unused.  Please remove this code example from the documentation.  log_error_write() is part of the lighttpd basic interfaces and is not specific to mod_openssl or wolfSSL.

gstrauss:
Please rebase dgarske:lighttpd_wolfssl against lighttpd:master and force push dgarske:lighttpd_wolfssl so that changes that I have made to lighttpd:master are not adding noise to the patches here.  Also, if you like, you may squash into the original commit the changes made to address my feedback.

src/mod_authn_file.c
TimParrish: MD4 is disabled in wolfSSL by default. However, MD4 is an available wolfSSL configure option, should the user desire it.

gstrauss:
Repeating what I posted 4 days ago:
Please rebase dgarske:lighttpd_wolfssl against lighttpd:master and force push dgarske:lighttpd_wolfssl so that changes that I have made to lighttpd:master are not adding noise to the patches here. Also, if you like, you may squash into the original commit the changes made to address my feedback.

Do not include patches which are unrelated to wolfSSL in this pull request.  Do not include patches related to compiling ./configure --without-pcre.  Those are completely unnecessary and a waste of both my time and your since I have already addressed these in lighttpd git master and have asked you to rebase your patch without them.

gstrauss:
The PK callbacks code looks to be removed in the latest push.  Was that intentional?

Please re-review the feedback I have submitted.  I hope I am misunderstanding you comment about "finalized this cleanup", since there are a number of remaining items from my feedback that need to be addressed, including the re-appearance of the private key in your most recent push.

gstrauss:
I updated lighttpd git master with support for wolfSSL, including build support for autoconf, meson, CMake, and SCons.

Remaining items for this pull request include only the doc/wolfssl contributions.  Please remove the private keys, remove the reference to wolfSSL_Debugging_ON() in the doc, remove trailing spaces in doc and other files.

RFE for wolfSSL: the wolfSSL openssl compatibility headers should all ```#include <wolfssl/options.h>``` instead of lighttpd having to do it.

gstrauss:
After further review of the remaining patch, I am going to close this pull request.  The gist is that wolfssl needs to be built with ./configure --enable-lighty, and then lighttpd needs to be built with ./configure --with-wolfssl

If you would like to add back code for the PK callbacks, please submit a more specific pull request.

If you post the documentation that you have written on the wolfssl.com website somewhere under support, then I will link to it from lighttpd's online documentation https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL

Thank you for your contribution.  Given the work put into the wolfSSL openssl-compatibility layer, the changes needed for lighttpd to use wolfSSL were quite small: e9f223d3  e1f21b2a  0074b6d3

dgarske:
Hi @gstrauss : I've captured this enhancement request internally, but feel free to submit a GitHub issue for it here: https://github.com/wolfSSL/wolfssl/issues

I agree it seems like inclusion of the wolfssl/options.h could happen in our library.

Thanks,
David Garske, wolfSSL

roytam1:
I wonder if I can use static wolfssl/wolfcrypt for lighttpd use? (my system in centos5 and I don't want to change system's openssl-0.9.8)

roytam1:
> ./configure --enable-lighty && make && sudo make install

alright, so having other `--enable-` {other-applications} causing compile failure.

gstrauss:
@roytam1 that is incomplete information and likely belongs as an issue on WolfSSL issue page rather than here.

dgarske:
Hi @roytam1 : It looks like there are some missing openssl compatiblity API's in wolfSSl with the latest Lighttpd. The last version we ported to and support is Lighttpd v1.4.51 on Aug 15, 2018 (a9e131fa376ade9aad7b7ee13098f6dd8337f84b)

Attached are the sources and patch for Lighttpd
[lighttpd-1.4.51.tar.gz](https://github.com/lighttpd/lighttpd1.4/files/2691862/lighttpd-1.4.51.tar.gz)
[Lighttpd_1_4_51.txt](https://github.com/lighttpd/lighttpd1.4/files/2691864/Lighttpd_1_4_51.txt)

See README in wolfStartup folder for details.

Thanks,
David Garske, wolfSSL

dgarske:
@roytam1 : Also it looks like `OPENSSL_NO_TLSEXT` should get defined to prevent the above errors with mod_openssl.c. On my builds that gets defined as the top of mod_openssl.c:27. If its not getting defined try adding `CFLAGS="-DOPENSSL_NO_TLSEXT"` to your Lighttpd `./configure.

Please make sure you are using the following steps:

wolfSSL:

```
./autogen.sh (if cloned from GitHub directly)
./configure --enable-lighty && make
sudo make install
sudo ldconfig
```

Lighttpd:
 
```
./autogen.sh (if cloned from GitHub directly)
./configure --with-wolfssl=yes
make
```

Thanks,
David Garske, wolfSSL

roytam1:
> Hi @roytam1 : It looks like there are some missing openssl compatiblity API's in wolfSSl with the latest Lighttpd. The last version we ported to and support is Lighttpd v1.4.51 on Aug 15, 2018 ([a9e131f](https://github.com/lighttpd/lighttpd1.4/commit/a9e131fa376ade9aad7b7ee13098f6dd8337f84b))

@dgarske would you please create new pull request for this?

dgarske:
@roytam1 : I've updated the wolfSSL to the latest Lighttpd commit in master a1077d18cb36b0cbdfb9da7138ba302f2090db78.

Many of the original fixes from this PR were already done in Lighttpd master. The changes are minimal and you can find the latest branch here:
https://github.com/dgarske/lighttpd1.4/tree/wolfssl_lighttpd

See the README in wolfStartup for details.

The steps above for building are still valid.

Thanks,
David Garske, wolfSSL

PR #93, opened 2018-11-10
[rand] Don't call RAND_cleanup with OpenSSL 1.1.x

neheb:
RAND_cleanup is deprecated and does nothing with 1.1.x.


gstrauss:
Please provide additional expanation in pull requests.

opensslconf-x86_64.h on my machines contains:
```
/*
 * Applications should use -DOPENSSL_API_COMPAT=<version> to suppress the
 * declarations of functions deprecated in or before <version>. Otherwise, they
 * still won't see them if the library has been built to disable deprecated
 * functions.
 */
```
and while OPENSSL_API_COMPAT defaults to 0, if the value is greater than or equal to 0x10100000L, and openssl 1.1.0 or later is used, then RAND_cleanup() is not defined.

lighttpd does not define OPENSSL_API_COMPAT, so this pull request is not fixing an issue, but is merely suggesting a tidy change.

lighttpd supports other TLS libraries with openssl-compatible interfaces, such as libreSSL, boringSSL, and wolfSSL.  How will they be impacted by the change in this pull request?

neheb:
I adjusted the commit message.

I also checked libressl's latest git. RAND_cleanup is present as an empty function. It is not guarded by anything.

BoringSSL is locked to 1.0.2 IIRC. wolfSSL seems to define 0x10001000L.

gstrauss:
BoringSSL master defines OPENSSL_VERSION_NUMBER 0x1010007f, different from what you guessed.  (Please be precise after being asked to provide clarification -- don't guess ("IIRC"))

LibreSSL and BoringSSL master branches both have RAND_cleanup() as empty routines.

wolfSSL does real work in RAND_cleanup(); it is not an empty routine in wolfSSL.

wolfSSL currently defines OPENSSL_VERSION_NUMBER 0x10001000L, as you have noted.  Additionally, wolfSSL openssl/opensslv.h has a comment suggesting it will increase OPENSSL_VERSION_NUMBER for HAVE_LIGHTY after compatibility for ECDH is added.

neheb:
From what I see, OPENSSL_VERSION_NUMBER is unreliable and only wolfSSL and OpenSSL <= 1.0.2 use the function.

Maybe a simple #ifdef could work.

neheb:
I edited the commit message slightly.

PR #94, opened 2018-12-04
[core] use kill_signal for gw_proc_kill

johannsong:
`kill-signal`, added in 1.4.15, isn't used with `idle-timeout` (which seems was originally added only for mod_scgi, and presently still uses SIGTERM via `gw_proc_kill`), therefore fastcgi processes couldn't be killed with `idle-timeout`.

### DIFF
--- a/src/gw_backend.c
+++ b/src/gw_backend.c
@@ -701,7 +701,7 @@ static void gw_proc_kill(server *srv, gw_host *host, gw_proc *proc) {
         host->unused_procs->prev = proc;
     host->unused_procs = proc;
 
-    kill(proc->pid, SIGTERM);
+    kill(proc->pid, host->kill_signal);
 
     gw_proc_set_state(host, proc, PROC_STATE_KILLED);
 


gstrauss:
> therefore fastcgi processes couldn't be killed with idle-timeout

The FastCGI documents SIGTERM to kill the process.  Looking at the code from a libfcgi fork I found, SIGUSR1 attempts a graceful shutdown, which does not work if the process if hung for any reason.

What are you expecting that you are saying that the fastcgi process couldn't be killed with idle-timeout?  Why isn't your FastCGI process responding to SIGTERM?  Please help me to understand so that the lighttpd behavior can be adjust appropriately.

johannsong:
Sorry about not being clear the first time. Neither SIGTERM nor SIGUSR1 works on the program (I was intending to just use SIGKILL instead). However, it looks like I (and my coworkers) have screwed up; this seems to be caused by the FastCGI version I'm using (long story).

In any case, the FastCGI code that I have is the [0910052249 snapshot](https://github.com/FastCGI-Archives/fcgi2/commit/fd11cd8d1291bab6d9bda91346c86746482b0b48); compared to 2.4.1 it had a new patch that sets the same SIGUSR1 handler for SIGTERM (unless one has already been set). The program had relied on the default SIGTERM behavior prior to that release (i.e. if the user doesn't explicitly install a handler then the default is used, which terminates the program immediately), whereas it appears the new "default" handler won't attempt to exit the program at all.

--------

Nevertheless, I think an option to kill the program with some other signal, e.g. SIGKILL, could still be helpful. The FastCGI spec does say to kill with SIGKILL if the program doesn't respond to SIGTERM, after all. I'm just not completely certain whether one `kill_signal` option is enough to cover all cases (e.g. when killing the process after idle-timeout, vs. when the server itself is exiting). (But *maybe* it's suggesting that we try SIGTERM first, but if the program is still alive after a while we should try to kill it again with SIGKILL.)

PR #96, opened 2019-02-05
doc/systemd: add example lighttpd.socket for systemd socket activation

glensc:
See discussion: https://github.com/lighttpd/lighttpd1.4/commit/ce7b47c015d6ce724ad6361c2df7a4853058654a#commitcomment-32195132

Systemd Docs:
- https://www.freedesktop.org/software/systemd/man/systemd.socket.html

However, not sure how to indicate that `server.systemd-socket-activation` needs to be enabled in config.

### DIFF
--- a/doc/config/lighttpd.conf
+++ b/doc/config/lighttpd.conf
@@ -104,6 +104,12 @@ server.use-ipv6 = "enable"
 server.username  = "lighttpd"
 server.groupname = "lighttpd"
 
+##
+## Enable lighttpd accepting connections from passed file descriptors.
+## https://www.freedesktop.org/software/systemd/man/systemd.socket.html
+##
+#server.systemd-socket-activation = "enable"
+
 ## 
 ## enable core files.
 ##
--- a/doc/systemd/Makefile.am
+++ b/doc/systemd/Makefile.am
@@ -1,1 +1,1 @@
-EXTRA_DIST=lighttpd.service
+EXTRA_DIST=lighttpd.service lighttpd.socket
--- a/doc/systemd/lighttpd.socket
+++ b/doc/systemd/lighttpd.socket
@@ -1,3 +1,6 @@
+# please note:
+# server.systemd-socket-activation option needs to be enabled in lighttpd config for socket activation to work.
+
 [Unit]
 Description=lighttpd socket
 
--- a/doc/systemd/lighttpd.socket
+++ b/doc/systemd/lighttpd.socket
@@ -1,12 +1,14 @@
-# please note:
-# server.systemd-socket-activation option needs to be enabled in lighttpd config for socket activation to work.
+# please note: lighttpd.conf must contain directive: server.systemd-socket-activation = "enable"
 
 [Unit]
 Description=lighttpd socket
 
 [Socket]
+# Enable listening on http port
 ListenStream=80
-ListenStream=443
+# To enable listening on https port, you need first setup SSL
+# https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL
+#ListenStream=443
 Service=lighttpd.service
 
 [Install]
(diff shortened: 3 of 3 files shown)


gstrauss:
> However, not sure how to indicate that `server.systemd-socket-activation` needs to be enabled in config.

Please add a comment to the lighttpd.socket file.  Lines that begin with '#' are comments.

glensc:
alright. done then.

NOTE: I have not tested this, don't have (working) systemd box around at this time.

gstrauss:
The sample lighttpd.conf in doc/config/lighttpd.conf does not have SSL enabled since certs would need to be configured.  Therefore, a sample lighttpd.socket should not enable listening on port 443 by default.  Please remove or comment out that line.

Please be more explicit in your comment:
    # please note: lighttpd.conf must contain directive: server.systemd-socket-activation = "enable"

In general, pull requests should be tested before being submitted.

glensc:
Here's some examples how other services use environment variables:
- https://freedesktop.org/wiki/Software/systemd/DaemonSocketActivation/

maybe something similar would not need the config directive.

why there's config option for now? because it's experimental? it imposes some security risk? performance overhead?

gstrauss:
One of the links you provided is to nginx and not appropriate for someone looking for basic documentation on socket activation.  (https://www.freedesktop.org/software/systemd/man/systemd.socket.html)

> why there's config option for now? because it's experimental? it imposes some security risk? performance overhead?

By default, lighttpd tries to trust the environment as little as possible.  If you want lighttpd to trust the environment for systemd socket activation on your specific machine with your specific configuration, then that is your choice and the config option is there to do so, but is disabled by default.

gstrauss:
The documentation: https://www.freedesktop.org/software/systemd/man/systemd.socket.html
e.g. a simple, user-friendly example: https://www.enricozini.org/blog/2017/debian/systemd-09-sockets/
though that second link is not unofficial, but still good and simple.

glensc:
Anything else left to do here?

I believe having this merged will urge downstream distributors to add support for socket activation in their distros.

gstrauss:
Yes.  There is more to do here.  Please re-read my two comments (8 Feb) which are **directly** above your most recent comment.

I disagree that this documentation, by itself, will "urge downstream distributors" do to anything.  One exception, of course, is the distribution to which you contribute.

glensc:
I don't understand, the only link to documentation in commits is https://www.freedesktop.org/software/systemd/man/systemd.socket.html

the nginx link is only in comments, not in commit.

you want the second unofficial link to be added in commit content?

by distributions, I mean if upstream includes sample `lighttpd.socket` in tarball, then downstream might consider installing it too.

gstrauss:
Yes, the link I questioned is in the comment above and not the comment.  Thank you for the correction.

doc/config/lighttpd.conf should add a comment line being explicit that systemd also needs to be configured, and a sample lighttpd.socket config can be found in doc/systemd/lighttpd.socket

doc/systemd/lighttpd.socket should be modified to be more explicit that setting up SSL has to be done in lighttpd.conf, e.g. "To enable listening on https port, you need first setup SSL in lighttpd.conf"

While systemd socket activation is quite simple, the samples are aimed at those who might not be familiar with socket activation, and perhaps are also not very familiar with lighttpd configuration.  Therefore, the documentation should aim to be simple, yet explicit.

gstrauss:
> doc/config/lighttpd.conf should add a comment line being explicit that systemd also needs to be
> configured, and a sample lighttpd.socket config can be found in doc/systemd/lighttpd.socket
> 
> doc/systemd/lighttpd.socket should be modified to be more explicit that setting up SSL has to be done
> in lighttpd.conf, e.g. "To enable listening on https port, you need first setup SSL in lighttpd.conf"

glensc:
@gstrauss thanks for the merge, sorry for not being able to address your notes myself.

PR #97, opened 2019-02-07
[mod_webdav] Various fixes

yangfl:
* Skip possible blank nodes
 * Iterate over all prop entries


gstrauss:
Having not looked at WebDAV for a while: are blank nodes valid?  If not, then why do they exist, and how did they get into the database?

Would you please give a concrete example of this extra level of properties?

yangfl:
See http://www.webdav.org/specs/rfc4918.html#rfc.section.9.2.2 . A newline is after <D:prop>, while the old code will treat that newline as a text node and insert it into db, with prop->name == "text". Also see http://xmlsoft.org/html/libxml-parser.html#xmlKeepBlanksDefault for its behavior on empty node.

yangfl:
Find prop with:
```bash
curl -X PROPFIND  -d '<D:propfind xmlns:D="DAV:" xmlns:N="EXAMPLE:"><D:prop><N:key/></D:prop></D:propfind>' http://localhost/webdav/file
```

OK:
```bash
curl -X PROPPATCH -d '<D:propertyupdate xmlns:D="DAV:" xmlns:N="EXAMPLE:"><D:set><D:prop><N:key>value</N:key></D:prop></D:set></D:propertyupdate>" http://localhost/webdav/file
```

BUG:
```bash
curl -X PROPPATCH -d '<D:propertyupdate xmlns:D="DAV:" xmlns:N="EXAMPLE:">
<D:set>
<D:prop>
<N:key>value</N:key>
</D:prop>
</D:set>
</D:propertyupdate>" http://localhost/webdav/file
```

gstrauss:
Thanks for the continued updates.  I have not forgotten about this, but it may take me a bit of time to come back to this.  I'll try to make some time this month to dig in deeper.

src/mod_webdav.c
```diff
@@ -34,7 +34,7 @@
 #endif
 
 #if defined(HAVE_LIBXML_H) && defined(HAVE_SQLITE3_H) \
- && defined(HAVE_UUID) && defined(HAVE_UUID_UUID_H)
+ && defined(HAVE_LIBUUID) && defined(HAVE_UUID_UUID_H)
```
gstrauss: configure.ac would need to be modified to define HAVE_LIBUUID, too, before making this change.

src/mod_webdav.c
```diff
@@ -869,8 +870,7 @@ static void webdav_activelock(buffer *b,
 	buffer_append_int(b, timeout);
 	buffer_append_string_len(b, CONST_STR_LEN("</D:timeout>\n"));
 
-	buffer_append_string_len(b, CONST_STR_LEN("<D:owner>"));
-	buffer_append_string_len(b, CONST_STR_LEN("</D:owner>\n"));
+	buffer_append_string(b, owner);
```
gstrauss: This looks incorrect.  Shouldn't the addition of owner be inbetween the <D:owner>...</D:owner> tags instead of removing those?

src/mod_webdav.c
```diff
@@ -1753,37 +1757,45 @@ static handler_t mod_webdav_put(server *srv, connection *con, plugin_data *p, ha
 		 * Example: Content-Range: bytes 100-1037/1038 */
 
 		if (NULL != (b = http_header_request_get(con, HTTP_HEADER_OTHER, CONST_STR_LEN("Content-Range")))) {
-			const char *num = b->ptr;
-			off_t offset;
+			char *num = b->ptr;
+			off_t offset, end, length;
```
gstrauss: must set default value for 'length', e.g. length = -1

src/mod_webdav.c
```diff
@@ -1753,37 +1757,45 @@ static handler_t mod_webdav_put(server *srv, connection *con, plugin_data *p, ha
...
-			while (*num == ' ' || *num == '\t') num++;
+				offset = strtoll(num, &err, 10);
+				if (num == err || *err != '-' || offset < 0) break;
+				num = err;
 
-			if (*num == '\0') {
-				con->http_status = 501; /* not implemented */
+				num++;
+				if (*num != '\0') {
+					end = strtoll(num, &err, 10);
```
gstrauss: please scope 'end' var to this block

src/mod_webdav.c
```diff
@@ -1753,37 +1757,45 @@ static handler_t mod_webdav_put(server *srv, connection *con, plugin_data *p, ha
...
-			offset = strtoll(num, &err, 10);
+					num++;
+					if (*num == '*') {
+						length = -1;
+					} else {
+						length = strtoll(num, &err, 10);
+						if (num == err || (length > 0 && length <= end)) break;
+					}
+				}
+				goto parse_success;
```
gstrauss: I am not a fan of goto.  Please replace the goto by setting err = NULL here on success, and error out below this block if (NULL != err)

src/mod_webdav.c
```diff
@@ -1796,6 +1808,16 @@ static handler_t mod_webdav_put(server *srv, connection *con, plugin_data *p, ha
 				return HANDLER_FINISHED;
 			}
 
+			if (length >= 0) {
```
gstrauss: Do not trust length from header blindly.

The uploaded content length (chunkqueue_length(cq)) (and cq is con->request_content_queue in this routine) should match the size of (end - offset) if end is set (scoped above), or (length - offset), and it should be an error if the length of content does not match.  (unless length = -1)

If end is provided and is different from length, then length should be checked against existing file size and it should be an error if the length does not match the current file length.

Since we are about to write the contents to the file, and I have added a requirement that length match the file size or about to be written to that file size, what is the benefit of the ftruncate()? 

(I am out of time at the moment.   I'll have to review the rest some other time.)

src/mod_webdav.c
```diff
@@ -1207,11 +1208,14 @@ static int webdav_parse_chunkqueue(server *srv, connection *con, handler_ctx *hc
...
 #ifdef USE_LOCKS
-static int webdav_lockdiscovery(connection *con, buffer *locktoken, const char *lockscope, const char *locktype, int depth) {
+static int webdav_lockdiscovery(connection *con, buffer *locktoken, const char *lockscope, const char *locktype, const char *owner, int depth) {
 
 	buffer *b = chunkqueue_append_buffer_open(con->write_queue);
 
-	http_header_response_set(con, HTTP_HEADER_OTHER, CONST_STR_LEN("Lock-Token"), CONST_BUF_LEN(locktoken));
+	/* wrap < > around the Lock-Token header */
+	char locktokenheader[buffer_string_length(locktoken) + 3];
+	snprintf(locktokenheader, sizeof(locktokenheader), "<%s>", locktoken->ptr);
```
gstrauss: Please do not use snprintf() for something so trivial.  snprintf() is slow.

Should the token be store in the database with the "<" ... ">"?  If not, then please (carefully) construct the string in the buffer before passing it to set the Lock-Token header value.

src/mod_webdav.c
yangfl: The pkg-config file for libuuid will take care of it, IIRC.

src/mod_webdav.c
yangfl: The format of owner in db is defined by https://github.com/lighttpd/lighttpd1.4/pull/97/files#diff-4c7084e5d651be053c7f6ac417f2fee0R2526 , with <D:owner>...</D:owner> wrap around the content of owner. This is because of the following example
```xml
<D:lockinfo xmlns:D='DAV:' xmlns:A='MY_NS'>
  <D:lockscope><D:exclusive/></D:lockscope>
  <D:locktype><D:write/></D:locktype>
  <D:owner>
    <D:href>http://example.org</D:href>
    <A:some-property>xxxx</A:some-property>
    plain text
  </D:owner>
</D:lockinfo>
```

There's another concern that values for "owner" are likely to be the same. Should they be stored in a separate table to reduce redundancy?

src/mod_webdav.c
yangfl: lighttpd (old version) does not store them with "<...>".

src/mod_webdav.c
yangfl: It does actually try to provide a way to truncate files remotely using the Content-Range scheme, but nah

src/mod_webdav.c
gstrauss: Your responses to my feedback do not give me confidence in your code.
You did not test this, and you are incorrect, at least on my system.

yangfl: It should build now

gstrauss: Thank you.

src/mod_webdav.c
```diff
@@ -1207,11 +1208,18 @@ static int webdav_parse_chunkqueue(server *srv, connection *con, handler_ctx *hc
...
+static int webdav_lockdiscovery(connection *con, buffer *locktoken, const char *lockscope, const char *locktype, const char *owner, int depth) {
 
 	buffer *b = chunkqueue_append_buffer_open(con->write_queue);
 
-	http_header_response_set(con, HTTP_HEADER_OTHER, CONST_STR_LEN("Lock-Token"), CONST_BUF_LEN(locktoken));
+	/* wrap < > around the Lock-Token header */
+	size_t locktokenlen = buffer_string_length(locktoken);
+	char locktokenheader[locktokenlen + 3];
+	locktokenheader[0] = '<';
+	strcpy(locktokenheader + 1, locktoken->ptr);
```
gstrauss: Better, but please never, never use strcpy().  Use memcpy() with locktokenlen to show that you know string length.  Also, the terminating '\0' is not required since the string length is passed to http_header_response_set().

src/mod_webdav.c
```diff
@@ -1750,33 +1758,51 @@ static handler_t mod_webdav_put(server *srv, connection *con, plugin_data *p, ha
...
-				con->http_status = 501; /* not implemented */
-
-				return HANDLER_FINISHED;
-			}
-
-			offset = strtoll(num, &err, 10);
-
-			if (*err != '-' || offset < 0) {
-				con->http_status = 501; /* not implemented */
+			/* TODO: if uploaded content length > request_length, the remain should be
```
gstrauss: Not a TODO.  con->request_content_queue contains exactly the request body contents.  mod_webdav waits until everything is uploaded and stores the request body in temporary files on local disk if the request body is large.

src/mod_webdav.c
yangfl: I think it's better to keep the terminating '\0' to avoid any possible buffer overread due to misuse.

gstrauss: The buffer is on the stack and is single use, calling a single API that was written to take a length so that substrings could be passed without terminating '\0'.  If you think you know better with your generic platitude, then you have not looked carefully at the code, which is well scoped, or at the API you are calling and into which you are passing that buffer.  On the other hand, you must have looked at the API to know whether or not you needed to pass in owned (allocated) memory or not, so your generic statement does not apply here.

If you wonder why I am so particular, please perform this test: benchmark your original snprintf() code 1 million times and compare it to copying into buffer with a few extra lines of code, also run 1 million times.  (In both cases, make sure that the optimizer does not optimize the loop away.)

src/mod_webdav.c
```diff
@@ -1792,20 +1809,56 @@ static handler_t mod_webdav_put(server *srv, connection *con, plugin_data *p, ha
...
 			}
 
-			if (-1 == lseek(fd, offset, SEEK_SET)) {
-				con->http_status = 501; /* not implemented */
+			do {
+				if (complete_length >= 0) {
+					off_t file_length = lseek(fd, 0, SEEK_END);
+					if (file_length < 0) break;
+					if (complete_length < file_length) {
+						/* RFC7233 Section 4.4 */
```
gstrauss: You mention RFC 7233 Section 4.4 here, but I do not think the changes in this commit are valid.  Would you please explain the problem you thought you were trying to solved with "improve Content-Range for PUT"?  What webdav client are you using which is sending Content-Range with PUT?  RFC 7231 rejects Content-Range with PUT.
Please see https://tools.ietf.org/html/rfc7231#section-4.3.4 :
>    An origin server that allows PUT on a given target resource MUST send
>    a 400 (Bad Request) response to a PUT request that contains a
>    Content-Range header field (Section 4.2 of [RFC7233]), since the
>    payload is likely to be partial content that has been mistakenly PUT
>    as a full representation.

yangfl: PUT with Content-Range is a popular extension for webdav since it enables editing a file just like doing it locally. See https://github.com/miquels/webdavfs/blob/master/DAV-requirements.md#for-writing-files , also Finder the OS X's built-in WebDAV client uses it.

yangfl: Since the client indicates the length of the resource, it must know it and if it's wrong, the server should warn the client instead of silently commiting changes. But if you are not satisfied with this part, I don't care about this particular behavior and it can be removed.

gstrauss: > But if you are not satisfied with this part

TBH, I am not satisfied with your cavalier approach to RFCs.  Please remove this patch, which is a **direct violation** of RFC7231.

https://tools.ietf.org/html/rfc7231#section-4.3.4 is very explicit about the required behavior.  Absent any errata that you can point me to which relaxes that requirement (and which I did not find), the MUST is a MUST.  That section in the RFC to which I linked follows with a pointers to how one might properly do a partial file update:
>    Partial content updates are possible by
>    targeting a separately identified resource with state that overlaps a
>    portion of the larger resource, or by using a different method that
>    has been specifically defined for partial updates (for example, the
>    PATCH method defined in [RFC5789]).

PATCH is not currently implemented in lighttpd mod_webdav and might be one proper approach to supporting this feature.  (I have a larger set of changes to mod_webdav which I plan to push in the next week or so, so if you choose to look into implementing PATCH, the base might change, so please hold off for now.)

yangfl: OK, I'll wait for those changes before I adjust my patches.

gstrauss:
I pushed a major rewrite of lighttpd mod_webdav.c (on a private branch from three years ago) to my development branch personal/gstrauss/master on git.lighttpd.net.  (https://redmine.lighttpd.net/projects/lighttpd/wiki/DevelGit)  Please consider it unstable at the moment.

You may wish to review it to see that many of your proposed changes are now much simpler:  It will be a one-line change to use xmlNodeIsBlank().  The Makefile changes to use HAVE_LIBUUID should still be relevant.  Wrapping the Lock-Token in "<...>" should be done carefully in the char array on the stack.  There is a new column in the database for storing ownerinfo.  It is obtained with xmlNodeListGetString(), though you might suggest the xmlNodeDump() that you used along with xmlBuffer.  Storing ownerinfo in a separate table (via foreign key) to reduce duplication sounds like a good idea for the future.

I see that the current (released) lighttpd mod_webdav.c violates the RFC by accepting Content-Range with PUT.  My rewrite from three years ago was more RFC compliant and removed that support.  I may consider re-adding support for compatibility with existing use, though I will continue to bemoan violating the RFCs.  Current limitation: lighttpd mod_webdav does not currently implement support for conditional requests (e.g. If-Match).  If lighttpd mod_webdav did so, then various partial PUT scenarios might be somewhat safer to implement in lighttpd mod_webdav, though proxies might still mishandle partial PUT.

yangfl:
Done. I left ownerinfo support behind since it needs major rewrite (to support foreign key and more).

gstrauss:
I cherry picked a couple patches, and made my own to surround Lock-Token with "<...>" without an additional copy.

While I could join tables to reduce the space taken by repeated ownerinfo strings, it does not seem worth the effort at this juncture, and would be additional effort to provide a migration path for existing database (which would have an different schema).  ownerinfo strings are not expected to be "too large". Even if each was 1k (quite large), then for 1000 files, that is 1MB, and for 10,000 files, that is 10MB.  While redundant, I do not feel them to be excessive enough to be an immediate problem for most common use cases.

Remaining questions for this pull request:
* xmlNodeListGetString() is used for ownerinfo.  Should it be changed to use xmlBufNodeDump()?
* how should lighttpd maintain compatibility with previous partial PUT with Content-Range?
  (incompatible behavior change introduced with my commits)

yangfl:
OK, I agree (separating ownerinfos might not be so helpful).

I didn't have a close look at new implementation regarding ownerinfo. Maybe I'll do it next week.

Maybe partial PUT with Content-Range could be treated just as new PATCH method, I guess.

gstrauss:
Your last push for namespace checks includes excess memcmp().  How does it handle less-compliant XML docs?  Is namespace always guaranteed to be set?  What to do if it is blank?

.

As for partial PUT, there are important design constraints for safety and corruption-avoidance, which I do not think has been acknowledged.  The mod_webdav rewrite that I did makes lighttpd much more compliant and safer to use.  Temporary files are renamed into place atomically.  Any requests receiving an older file should receive the complete older file, no matter how slow it is downloading.  New requests should receive the new file.

If partial PUT is reintroduced, then how will concurrent GET requests see a view of either the complete old file or the complete new file?  (and send a corresponding etag?)  How will race conditions be avoided?

I have considered multiple approaches, and there are tradeoff to be made, or limitations to be placed, on various alternatives.

In short, there is no simple, immediate answer, so please think carefully about this before investing time in a solution which might be rejected.

gstrauss:
I accepted a few small patches from this pull request.

> Maybe partial PUT with Content-Range could be treated just as new PATCH method, I guess.

Please do not guess.  I am not interested in uninformed guesses.

In the future, please keep your pull requests small and focused.  Test them extensively and provide references to RFCs or other reputable sources to justify the changes you are proposing.  Thank you.

src/mod_webdav.c
```diff
@@ -256,8 +255,10 @@ int mod_webdav_plugin_init(plugin *p) {
 #define WEBDAV_FLAG_MOVE_XDEV    0x10
 #define WEBDAV_FLAG_COPY_XDEV    0x20
 
-#define webdav_xmlstrcmp_fixed(s, fixed) \
-        memcmp((s), (fixed), sizeof(fixed))
+#define webdav_xmlnode_isdav(prop, attr) \
+        ((prop)->ns && \
+            0 == memcmp((prop)->ns->href, "DAV:", sizeof("DAV:")) && \
+            0 == memcmp((prop)->name, (attr), sizeof(attr)))
```
gstrauss: If considering this for the future, put the memcmp() of prop->name first, to short-circuit the checks early if not a match.

PR #98, opened 2019-02-22
Fix compile error when system has SOCK_CLOEXEC but not SOCK_NONBLOCK

supergaute:
My system has SOCK_CLOEXEC but not SOCK_NONBLOCK. This cause the build to break.
This commit should fix this.

### DIFF
--- a/src/fdevent.c
+++ b/src/fdevent.c
@@ -165,10 +165,16 @@ fdevents *fdevent_init(server *srv) {
 	 * (reported on Android running a custom ROM)
 	 * https://redmine.lighttpd.net/issues/2883
 	 */
+	#ifdef SOCK_NONBLOCK
 	int fd = socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC | SOCK_NONBLOCK, 0);
+	#else
+	int fd = socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0);
+	#endif
 	if (fd >= 0) {
+		#ifdef SOCK_NONBLOCK
 		int flags = fcntl(fd, F_GETFL, 0);
 		use_sock_nonblock = (-1 != flags && (flags & O_NONBLOCK));
+		#endif
 		use_sock_cloexec = 1;
 		close(fd);
 	}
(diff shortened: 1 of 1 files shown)


gstrauss:
SOCK_NONBLOCK is expected to exist if SOCK_CLOEXEC exists, and we even test if SOCK_NONBLOCK is functional (see the comment directly under your first patch).  What system are you on where that is not the case?  (and do you have any idea what kinds of security holes are in that kernel?)

gstrauss:
SOCK_NONBLOCK and SOCK_CLOEXEC were both added to Linux in kernel 2.6.27.  Please provide more details as to what OS system (and version) you are on that you claim has SOCK_CLOEXEC but not SOCK_NONBLOCK.

supergaute:
This is on a QNX 7.0 system.

gstrauss:
> This is on a QNX 7.0 system.

That is a relevant piece of information that should not have been omitted.

Why not configure the code to use SOCK_CLOEXEC where the flag is available?
```
--- a/src/fdevent.c
+++ b/src/fdevent.c
@@ -165,10 +165,16 @@ fdevents *fdevent_init(server *srv) {
         * (reported on Android running a custom ROM)
         * https://redmine.lighttpd.net/issues/2883
         */
+       #ifdef SOCK_NONBLOCK
        int fd = socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC | SOCK_NONBLOCK, 0);
+       #else
+       int fd = socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0);
+       #endif
        if (fd >= 0) {
                int flags = fcntl(fd, F_GETFL, 0);
+              #ifdef SOCK_NONBLOCK
                use_sock_nonblock = (-1 != flags && (flags & O_NONBLOCK));
+              #endif
                use_sock_cloexec = 1;
                close(fd);
        }
@@ -454,9 +460,20 @@ int fdevent_socket_cloexec(int domain, int type, int protocol) {
 int fdevent_socket_nb_cloexec(int domain, int type, int protocol) {
        int fd;
 #ifdef SOCK_CLOEXEC
+#ifdef SOCK_NONBLOCK
        if (use_sock_cloexec && use_sock_nonblock)
                return socket(domain, type | SOCK_CLOEXEC | SOCK_NONBLOCK, protocol);
+#else
+       if (use_sock_cloexec
+           && -1 != (fd = socket(domain, type | SOCK_CLOEXEC, protocol))) {
+#ifdef O_NONBLOCK
+               force_assert(-1 != fcntl(fd, F_SETFL, O_NONBLOCK | O_RDWR));
+#endif
+               return fd;
+       }
 #endif
+#endif
+
        if (-1 != (fd = socket(domain, type, protocol))) {
 #ifdef FD_CLOEXEC
                force_assert(-1 != fcntl(fd, F_SETFD, FD_CLOEXEC));
```

supergaute:
This approach works fine on my system...

supergaute:
Any updates on this?

PR #99, opened 2019-02-27
[mod_skeleton] cleaner calloc use in SETDEFAULTS_FUNC

pksadiq:


gstrauss:
How about "cleaner calloc() usage in module SETDEFAULTS_FUNC"

Also, the syntax change might trigger warnings in some static analyzers that think that srv->config_context->used might theoretically be 0, even though it will not be since there should always be a global context.

pksadiq:
> Also, (untested) the syntax change might trigger warnings in some static analyzers that think that srv->config_context->used might theoretically be 0

It's legal to pass 0 as the first argument to calloc, which will simply return a NULL. So that shouldn't be a concern

gstrauss:
"[multiple] cleaner calloc use in SETDEFAULTS_FUNC"
fits in 50 chars, so I made one more edit to title.

gstrauss:
FYI: Looks like the current calloc usage syntax was introduced in commit bac9212c8 a long, long time ago.

yangfl:
Look like there're much more https://github.com/lighttpd/lighttpd1.4/search?q=calloc%281%2C+srv-%3Econfig_context-%3Eused&unscoped_q=calloc%281%2C+srv-%3Econfig_context-%3Eused

gstrauss:
If you would like to extend this patch to cover the calloc() usage in SETDEFAULT_FUNC in all the modules, I'll review that, the result of your scripted replacement.  Thanks!

gstrauss:
Thanks for the update.  Sorry, I must have missed the email notification.  I plan to review the patch this weekend, if not sooner.

PR #100, opened 2019-10-10
[systemd]: update /var/run -> /run for systemd

glensc:
This gets rid of the warning:
> May 19 10:56:32 buster systemd[1]: /lib/systemd/system/lighttpd.service:6: PIDFile= references path below legacy directory /var/run/, updating /var/run/lighttpd.pid → /run/lighttpd.pid; please update the unit file accordingly.

refs:
- https://github.com/systemd/systemd/commit/a2d1fb882c4308bc10362d971f333c5031d60069
- https://github.com/systemd/systemd/pull/9019
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=929203

### DIFF
--- a/doc/systemd/lighttpd.service
+++ b/doc/systemd/lighttpd.service
@@ -4,7 +4,7 @@ After=network-online.target
 
 [Service]
 Type=simple
-PIDFile=/var/run/lighttpd.pid
+PIDFile=/run/lighttpd.pid
 ExecStartPre=/usr/sbin/lighttpd -tt -f /etc/lighttpd/lighttpd.conf
 ExecStart=/usr/sbin/lighttpd -D -f /etc/lighttpd/lighttpd.conf
 ExecReload=/bin/kill -USR1 $MAINPID


gstrauss:
Thank you for the patch.  Please submit a patch to salsa.debian.org, and include changes to all locations that reference /run, including doc/lighttpd.8 and doc/config/lighttpd.conf

glensc:
I don't get it, is it rejected here? Or why must I submit this to Debian bug tracker?

In some other [thread](https://bbs.archlinux.org/viewtopic.php?id=243455) (archlinux) they said this is not package maintainers to fix, but upstream, which makes sense to cut down the duplication.

I agree that consistency has value, perhaps the `/var/run` should be replaced with `@placeholders@` so downstream distributions can use `./configure` argument?

ps: I'm not using this package on Debian myself.

gstrauss:

mckaygerhard:
why is systemd forced? .. not all the systems used systemd you know?

PR #101, opened 2019-11-13
Solaris has getloadavg in sys/loadavg.h

niclasr:
SunOS/Solaris/Illumos has getloadavg declared in sys/loadavg.h .
Lighttpd uses getloadavg but does not include the header that
declares it on SunOs/Solaris/illumos. This patch fixes that.


gstrauss:
Thanks @niclasr 
Will review.

gstrauss:

PR #102, opened 2020-01-30
[PATCH] Fix build when using nested CMake

raanwi:
This fixes the CMake build when nesting it via FetchContent or ExternalProject_Add. Makes life a lot easier for distributors.


gstrauss:

PR #103, opened 2020-04-09
malloc without free

klebertarcisio:
Hi everyone,

I found an opportunity for improvement this code. I believe in this case we need to call the free() function. In this case, we have 2 types of weaknesses: CWE 399 (Resource Management Errors) and CWE 772 (Missing Release of Resource after Effective Lifetime).

According to CWE 772: "When a resource is not released after use, it can allow attackers to cause a denial of service by causing the allocation of resources without triggering their release. Frequently-affected resources include memory, CPU, disk space, power or battery, etc."

CWE 339: https://cwe.mitre.org/data/definitions/399.html
CWE 772: https://cwe.mitre.org/data/definitions/772.html

Regards

### DIFF
--- a/src/lemon.c
+++ b/src/lemon.c
@@ -3869,6 +3869,7 @@ char *data;
     free(x1a->tbl);
     /* *x1a = array; *//* copy 'array' */
     memcpy(x1a, &array, sizeof(array));
+    free(array.tbl);	  
   }
   /* Insert the new data */
   h = ph & (x1a->size-1);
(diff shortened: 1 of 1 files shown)


stbuehler:
`lemon` is the parser generator, run at build time. nobody cares.

klebertarcisio:
Thanks for the answer. In the lemon file, we have 18 calls to the free () function. In addition, we have 9 calls to the fclose () function. I believe it must be important to free up resources when we no longer want to use them.

gstrauss:
@stbuehler wrote: `lemon` is the parser generator, run at build time.

As such, there is no external exposure to unknown attackers.  The nature of the `lemon` parser is that it is a single-use short-run program, and used only at build time in lighttpd.  It is not a long-lived daemon.

@klebertarcisio the upstream lemon parser can be found at
https://en.wikipedia.org/wiki/Lemon_(parser_generator)
https://www.sqlite.org/src/file/tool/lemon.c
You might look to see if your patches apply to the maintained version in SQLite.

PR #105, opened 2021-05-15
Fix kill children and shutdown by signal

pgul:
Graceful shutdown by signal to master process was broken in 29f50bde.
We should not quiet ignore EINTR in the main loop.

### DIFF
--- a/src/fdevent.c
+++ b/src/fdevent.c
@@ -815,6 +815,10 @@ int fdevent_waitpid(pid_t pid, int * const status, int nb) {
     return rv;
 }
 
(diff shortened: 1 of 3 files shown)


gstrauss:
Thanks.

`fdevent_waitpid()` was created so that wait interfaces could be more isolated, so I would prefer an addition interface for an interruptable `wait()`, e.g. `fdevent_waitpid_intr()`.

gstrauss:
Thanks!  I've merged your patch into my development branch and it will be on lighttpd git master in a few days.

PR #107, opened 2021-07-27
[wolfssl] Correct return check for `wolfSSL_sk_X509_NAME_push`

julek-wolfssl:
We received a report about an error loading CA certs in lighttpd when using wolfSSL. This patch corrects the return code checked by lighttpd.

Building lighttpd, I found that wolfSSL includes in `src/sys-crypto-md.h` were missing the build configuration.

### DIFF
--- a/src/mod_wolfssl.c
+++ b/src/mod_wolfssl.c
@@ -790,6 +790,10 @@ mod_wolfssl_CTX_use_certificate_chain_file (WOLFSSL_CTX *ssl_ctx, const char *fn
 static STACK_OF(X509_NAME) *
 mod_wolfssl_load_client_CA_file (const buffer *ssl_ca_file, log_error_st *errh)
 {
+    /* wolfSSL_sk_X509_NAME_push prior to version 4.8.0 returned 0
+     * on success. This was changed to WOLFSSL_SUCCESS from 4.8.0. */
+    int wolf_success_rc = wolfSSL_lib_version_hex() < 0x04008000
+                            ? 0 : WOLFSSL_SUCCESS;
     /* similar to wolfSSL_load_client_CA_file(), plus some processing */
     buffer **certs = NULL;
     if (NULL == mod_wolfssl_load_pem_file(ssl_ca_file->ptr, errh, &certs)) {
@@ -814,7 +818,7 @@ mod_wolfssl_load_client_CA_file (const buffer *ssl_ca_file, log_error_st *errh)
         WOLFSSL_X509_NAME *subj = NULL;
         if (NULL == ca
             || NULL == (subj = wolfSSL_X509_get_subject_name(ca))
-            || 0 != wolfSSL_sk_X509_NAME_push(canames,
+            || wolf_success_rc != wolfSSL_sk_X509_NAME_push(canames,
                                               wolfSSL_X509_NAME_dup(subj))) {
             log_error(errh, __FILE__, __LINE__,
               "SSL: couldn't read X509 certificates from '%s'",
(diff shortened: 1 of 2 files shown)


gstrauss:
Your patch suggests a logic inversion error checking the return value of `wolfSSL_sk_X509_NAME_push()`.  However, my read of the wolfssl code shows that 0 returned from `wolfSSL_sk_X509_NAME_push()` is success.  Please help me to understand why you think this is a logic inversion error.

Separately, I have questions about the need for your change to `src/sys-crypto-md.h`.  Please see 8187e988

gstrauss:
I do see that 8187e98 removed `<wolfssl/options.h>` from `src/sys-crypto.h` and so in `src/sys-crypto-md.h`, any wolfssl `NO_*` defines are not present.  I need to check further to see if there are unwanted interactions between `<wolfssl/options.h>` included in `src/sys-crypto-md.h` when lighttpd is built with multiple TLS library options enabled (e.g. openssl, gnutls, mbedtls, nss, wolfssl)

julek-wolfssl:
`wolfSSL_sk_X509_NAME_push` returns `WOLFSSL_SUCCESS`(1) on success while `WOLFSSL_FAILURE`(0) is returned on failure.

Regarding the changes in `src/sys-crypto-md.h`, I got errors compiling the master branches of both projects.

gstrauss:
> wolfSSL_sk_X509_NAME_push returns WOLFSSL_SUCCESS(1) on success while WOLFSSL_FAILURE(0) is returned on failure.

I see that you made a logical-inversion of `wolfSSL_sk_X509_NAME_push()` return value in https://github.com/wolfSSL/wolfssl/commit/1b6b16c2c39796e73262befe89c6abc3d6cf770f
Do you realize that is incompatible with previous versions of wolfssl?

julek-wolfssl:
`wolfSSL_sk_X509_NAME_push` is part of the OpenSSL compatibility layer. We want for the API return code to be compatible with OpenSSL. The return code was incompatible and needed to be changed. I realize this breaks backwards compatibility with wolfSSL but it is more important to correct wrong functionality.

gstrauss:
> wolfSSL_sk_X509_NAME_push is part of the OpenSSL compatibility layer. We want for the API return code to be compatible with OpenSSL. The return code was incompatible and needed to be changed. I realize this breaks backwards compatibility with wolfSSL but it is more important to correct wrong functionality.

Does WolfSSL support older versions of the wolfssl libraries?
Did WolfSSL v4.8.0 release notes highlight this intentional incompatibility with earlier versions?
If this change was intentional, why did you not highlight this information in this PR?  Why did you make me, an independent developer, dig through wolfssl code to find the breakage was caused by a wolfssl code change?  Did you tell the customer that this is a bug in lighttpd, or did you tell the customer that you intentionally changed the interface for `wolfSSL_sk_X509_NAME_push()` and broke compatibility, which lighttpd could not have foreseen.  lighttpd code using `wolfSSL_sk_X509_NAME_push()` is correct for wolfssl v4.7.1r and earlier.

lighttpd strives to work with libraries that are part of (modern releases of) numerous OS distributions.  Many of these distros are not running the very latest release of wolfssl.  Would you please modify your patch to check the wolfssl version against which lighttpd is being compiled, and to choose the appropriate success values against which to check the return code from `wolfSSL_sk_X509_NAME_push()`?

I suggest something like
```
#if LIBWOLFSSL_VERSION_HEX < 0x04008000
int wrc = 0;
#else
int wrc = WOLFSSL_SUCCESS;
#endif
```
and then comparing each call to `wolfSSL_sk_X509_NAME_push()` to `wrc`.

gstrauss:
`wolfSSL_sk_X509_NAME_push()` is used in lighttpd at startup, so a runtime check of the wolfssl library version is also an option, and more compatible.

julek-wolfssl:
It was determined to be an incorrect usage due to a recent change in the function. I understand your frustration with this change and I admit that we should have communicated this better. This one is on us. I will correct this PR with your runtime check suggestion.

julek-wolfssl:
I pushed a correction to check the return code at run-time. If you find a discrepancy between our compatibility API and OpenSSL's, please contact us so that we may fix it and document it correctly.

gstrauss:
Thank you for updating the patch.  I have committed a slightly modified version to my development branch, keeping you as the committer and adding info to the commit message.

This PR will be closed automatically once my development branch is pushed to lighttpd git master, which I expect to do later in August.

julek-wolfssl:
Thanks!

PR #108, opened 2021-09-20
Fix config files

istiak101:
Fix default lighttpd.conf file and add vhost level error log example in vhosts.template.


gstrauss:
Thanks!  Your updates make sense.  I'll include this patch in the next batch I push.

istiak101:
Thanks!

PR #109, opened 2021-10-13
[core] forward connection reset from backend to client

antosip:
Let's add option to reset connection when backend connection is reset.
Useful when need to abort file download with failure indication (e.g. in web browser).

### DIFF
--- a/src/base.h
+++ b/src/base.h
@@ -120,6 +120,7 @@ typedef struct {
 	unsigned char config_unsupported;
 	unsigned char systemd_socket_activation;
 	unsigned char errorlog_use_syslog;
+	unsigned char forward_reset;
 	const buffer *syslog_facility;
 	const buffer *bindhost;
 	const buffer *changeroot;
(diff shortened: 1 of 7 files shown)


gstrauss:
> I see no other way to communicate error back to client other than connection reset. May be you know any?

Modern versions of lighttpd mod_proxy use HTTP/1.1 to connect to backend.  If your backend returns `Transfer-Encoding: chunked`, and uses chunked encoding on the response, then lighttpd will return the same to HTTP/1.1 clients, and those clients can detect a truncated response that did not receive an end-chunk.

gstrauss:
> I see no other way to communicate error back to client other than connection reset. May be you know any?

Answered above.  HTTP/1.1 clients will receive `Transfer-Encoding: chunked` for a streamed response from the backend (when lighttpd.conf is configured with `server.stream-response-body = 2`), and the clients should be able to detect a truncated response.

PR #110, opened 2021-11-04
Config file update

istiak101:
Removed socket example from vhosts.template as that is already covered in lighttpd.conf. Added conditional https scheme instead.

### DIFF
--- a/doc/config/modules.conf
+++ b/doc/config/modules.conf
@@ -51,7 +51,6 @@
 server.modules = (
 #  "mod_rewrite",
   "mod_access",
-#  "mod_openssl",
 #  "mod_evasive",
 #  "mod_auth",
 #  "mod_authn_file",
--- a/doc/config/modules.conf
+++ b/doc/config/modules.conf
@@ -51,6 +51,7 @@
 server.modules = (
 #  "mod_rewrite",
   "mod_access",
+#  "mod_openssl",
 #  "mod_evasive",
 #  "mod_auth",
 #  "mod_authn_file",
@@ -70,57 +71,57 @@ server.modules = (
 ##
 ## mod_expire
 ##
-#include "conf.d/expire.conf"
+#include conf_dir + "/conf.d/expire.conf"
 
 ##
 ## mod_deflate
 ##
-#include "conf.d/deflate.conf"
+#include conf_dir + "/conf.d/deflate.conf"
 
 ##
 ## mod_magnet
 ##
-#include "conf.d/magnet.conf"
+#include conf_dir + "/conf.d/magnet.conf"
 
 ##
 ## mod_geoip
 ##
-#include "conf.d/geoip.conf"
+#include conf_dir + "/conf.d/geoip.conf"
 
 ##
 ## mod_ssi
 ##
-#include "conf.d/ssi.conf"
+#include conf_dir + "/conf.d/ssi.conf"
 
 ##
 ## mod_status
 ##
-#include "conf.d/status.conf"
+#include conf_dir + "/conf.d/status.conf"
 
 ##
 ## mod_webdav
 ##
-#include "conf.d/webdav.conf"
+#include conf_dir + "/conf.d/webdav.conf"
 
 ##
 ## mod_userdir
 ##
-#include "conf.d/userdir.conf"
+#include conf_dir + "/conf.d/userdir.conf"
 
 ##
 ## mod_cml
 ##
-#include "conf.d/cml.conf"
+#include conf_dir + "/conf.d/cml.conf"
 
 ##
 ## mod_rrdtool
 ##
-#include "conf.d/rrdtool.conf"
+#include conf_dir + "/conf.d/rrdtool.conf"
 
 ##
 ## mod_secdownload
 ##
-#include "conf.d/secdownload.conf"
+#include conf_dir + "/conf.d/secdownload.conf"
 
 ##
 #######################################################################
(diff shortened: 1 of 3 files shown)


doc/config/modules.conf
```diff
@@ -51,6 +51,7 @@
 server.modules = (
 #  "mod_rewrite",
   "mod_access",
+#  "mod_openssl",
```
gstrauss, line 54: I do not think maintaining a complete list here is necessary.

Besides, lighttpd supports multiple TLS modules as options.  If one is listed, then all should be listed.  Instead, I think none should be listed here.  On the other hand, if there are uncommented `ssl.*` directives in the examples, there should probably be an uncommented `server.modules` including a TLS module.

lighttpd supports and ignores duplicated modules in `server.modules` since lighttpd 1.4.56, so this can be located near examples: `server.modules += ( "mod_openssl" )`.  `server.modules += ("mod_openssl")` should be added to the commented out example in the sample `lighttpd.conf`.

doc/config/vhosts.d/vhosts.template
```diff
@@ -20,14 +20,15 @@ $HTTP["host"] == "download.example.com" {
   ##
   accesslog.filename          = log_root + "/" + server_name + "/access.log"
   server.errorlog             = log_root + "/" + server_name + "/error.log"
-}
 
-$SERVER["socket"] == "127.0.0.1:443" {
-  server.name = "localhost"
-  ssl.pemfile = "/etc/ssl/private/lighttpd-localhost.pem"
-  ssl.engine  = "enable"
+  $HTTP["scheme"] == "https" {
```
gstrauss, line 24: Your suggested change appears to me to be to be redundant and possibly misleading.  If you have `ssl.engine = "enable"`, the scheme for requests should be `https`.  It is an error in lighttpd not to have a default certificate defined for a `$SERVER["socket"]` when `ssl.engine = "enable"` elsewhere in the config.  `ssl.engine = "enable"` should only appear in the global scope or in `$SERVER["socket"]`, and `$SERVER["socket"]` should not be nested inside other lighttpd conditions.

Related to my comment at the top of the review, the (original) example should probably be commented out, or else `server.modules += ( "mod_openssl" )` should be present.

On second thought, the original ssl example here was to illustrate that an alternative document root could be used, but I think this ssl section here should be removed entirely.

gstrauss (changes requested):
Thank you for your suggestions.
```
#include "conf.d/expire.conf"
#include conf_dir + "/conf.d/expire.conf"
```
These (and related) are acceptable changes in this PR.  I left comments by the other changes, and suggested they be removed from this PR.

doc/config/modules.conf
istiak101: I have added `server.modules += ( "mod_openssl" )` in the commented out section of sample `lighttpd.conf` as per your suggestion and removed `mod_openssl` from `modules.conf`.

doc/config/vhosts.d/vhosts.template
istiak101: Removed `ssl.engine = "enable"` from here and commented out rest of the example.

doc/config/vhosts.d/vhosts.template
```diff
@@ -21,14 +21,13 @@ $HTTP["host"] == "download.example.com" {
...
-    server.name = "download.example.com"
-    ssl.pemfile = "/etc/lighttpd/certs/download.example.com.cert.pem"
-    ssl.privkey = "/etc/lighttpd/certs/download.example.com.key.pem"
-    ssl.engine  = "enable"
+#  $HTTP["scheme"] == "https" {
+#    server.name = "download.example.com"
+#    ssl.pemfile = "/etc/lighttpd/certs/download.example.com.cert.pem"
+#    ssl.privkey = "/etc/lighttpd/certs/download.example.com.key.pem"
 
-    server.document-root = vhosts_dir + "/example.com/download/htdocs"
```
gstrauss: Why should this section be here at all?  I already questioned that `$HTTP["scheme"]` is redundant, and you have not addressed that.  What value is changing the server.document-root for https, and why does it belong in vhosts.template?  Keep in mind that the historical doc was written long ago, and I am questioning whether or not it is still relevant today, when most sites are using TLS.

doc/config/vhosts.d/vhosts.template
gstrauss: I would like to remove the entire section.  You may do so as part of this patch, or I can do so separately.  If you are making changes, such as commenting the section out, please either justify why you are making changes here, or revert your changes here.

doc/config/vhosts.d/vhosts.template
istiak101: Should I remove this completely or just the `server.document-root` line? I kept this as an example. I should've changed the value.

doc/config/vhosts.d/vhosts.template
istiak101: I added this conditional way as example to easily have different certificates per vhost. I know there are other possible ways to implement this. I commented this out because as you stated that if turn on ssl.engine I must have a default certificate for the server.

gstrauss: I wrote:
> On second thought, the original ssl example here was to illustrate that an alternative document root could be used, but I think this ssl section here should be removed entirely.

What was unclear?

You wrote:
> I added this conditional way as example to easily have different certificates per vhost. I know there are other possible ways to implement this.

The only relevant way to choose certificates based on vhost is `$HTTP["host"]`.  Every other way you can think of, including "scheme" is not based on vhost and therefore does not belong with the vhost config example.

istiak101: Sorry, my bad. I have removed that part completely.

doc/config/vhosts.d/vhosts.template
gstrauss: Since you have not provided a reasoning why you desire changes here, you should remove all of your changes here, in this part of the patch.  Since I have presented a reason why I plan to make changes, I will (separately) make the change for which I have shared my reasoning.  Put another way, when you are making contributions, please carefully scope your changes to changes that you are comfortable explaining and justifying.

istiak101: Sure, thanks.

gstrauss:
Patches merged.  Thank you.

FYI: I pushed a subsequent commit to `vhosts.template` with sample TLS config, commented out.

istiak101:
Thank you very much. Looks good!

PR #116, opened 2022-09-20
meson fixes

neheb:
This bumps the required version from seemingly 0.40.0 to 0.47.0 to take advantage of feature options.

libxml2 does not work as a wrap. Still need to investigate what's wrong with it.


gstrauss:
Thanks for putting this together!

I need to learn more about these features of meson...

Minor nit: I'd prefer to check if an option is set before doing the work to check if the dependency exists.
The meson setup is faster if we skip searching for dependencies that are not going to be used.
However, the objects will still need to be initialized before later use such as `libfoo.found()`

I have a concern about maintaining the subprojects.  The benefit of pkgconfig is that the packages provide a .pc for themselves.  On the other hand, these subpackage definitions duplicate a lot of information about (other) subpackages.  I'd prefer to simply provide a minimum required version, e.g. similar to meson_version, or similar to debian/control Build-Depends in the lighttpd Debian package.

neheb:
not sure what you mean. The whole point of features is to reduce codesize. eg:

meson -Diconv=disabled
dependency('iconv', required: get_option('iconv')) <-- will be not found even if available, which it is on most OSes.

the subprojects come from WrapDB. They get installed with

```
meson wrap install X
```

https://github.com/mesonbuild/wrapdb is where they're maintained. Note that the dependency typos I was able to find by doing

```
meson -Dwrap_mode=forcefallback
```

and noticing that they were not being used.

Custom subprojects can be created but I doubt there's much reason to.

I added pcre as there's code for it. Removing pcre support is beyond the scope of this PR.

As an aside, does lighttpd really support windows? Last I tested it failed to compile.

eli-schwartz:
> Minor nit: I'd prefer to check if an option is set before doing the work to check if the dependency exists.
> The meson setup is faster if we skip searching for dependencies that are not going to be used.
> However, the objects will still need to be initialized before later use such as `libfoo.found()`

This is actually why you really want feature-keyed dependencies. A single dependency lookup can log:
> "Dependency **libfoo** skipped: feature **foo** disabled"

and will not try to run pkg-config, and the object is initialized as a not-found-dependency.

Since you pass the *feature* option as an argument to `dependency()`, the dependency function itself is able to internally check if the option is set.

gstrauss:
Thanks @eli-schwartz .  I was not aware that meson `dependency()` would short-circuit its work that for 'feature'.  The stock `dependency()` will look for the object and provide an object if found, but will not throw an exception if the dependency is not found and `required: false`.

> As an aside, does lighttpd really support windows? Last I tested it failed to compile.

lighttpd is supported under cygwin on Windows.

(I have an unreleased experimental branch for Windows, but the Windows API is horrid.  In some 40+ years, Microsoft could have had interns easily provide many, many POSIX compatibility wrappers (though not everything) and Microsoft has failed to make any convincing effort to do so, e.g. `close()` on sockets, `getopt()`, `opendir()`, `readdir()`, `closedir()`, `socketpair()`, `stat()`, `gmtime_r()`, `localtime_r()`, `glob()`,  and more that I have had to write myself using often-inferior Windows APIs.  ...Maybe I'll toy around with a meson static build on Windows with my experimental branch.)

---

@neheb 
I'll try to experiment with this in the next few days so that I can get a better understanding of the maintenance expectations.

neheb:
No way it compiles on windows. As an example, the cmake and meson files reference an xgetopt.c file that's not even there.

gstrauss:
> No way it compiles on windows. As an example, the cmake and meson files reference an xgetopt.c file that's not even there.

Hahaha.  Yes.  That's old.

If you're feeling masochistic, I rebased and pushed my win32-exp branch to https://github.com/gstrauss/lighttpd1.4/tree/win32-exp but I have not tried to compile it in nearly a year, and it does not contain whatever fixes I was in the middle of working on in that tree on Windows, as that work in progress is on a bastion Windows host I almost never boot into.

eli-schwartz:
> Thanks @eli-schwartz . I was not aware that meson `dependency()` would short-circuit its work that for 'feature'. The stock `dependency()` will look for the object and provide an object if found, but will not throw an exception if the dependency is not found and `required: false`.

Yup, that corresponds to the user defining the option as `=auto` -- it will look for the object, and provide it if possible, but not throw an error if it is not-found.

Autotools convention would be to call this `--with-foo=check`.

In general, you want to encode 3 `required: ` states:
- feature `-Dfoo=enabled` == required, fail if not found, equivalent to boolean `-Dfoo=true`, autotools `--with-foo=yes`
- feature `-Dfoo=auto` == automagic dependency, only use if found, equivalent to boolean `-Dfoo=false`, autotools `--with-foo=check`
- feature `-Dfoo=disabled` == immediately skip, equivalent to if/else statements that avoid running `dependency()`, but with better logging, autotools `--with-foo=no`

When using classic booleans instead of neat new features, it was traditionally necessary to choose between options 2 and 3 when handling boolean false, by either going with option 2 and directly passing the option to the dependency, or choosing option 3 and complicating the build description with if/else code.

gstrauss:
@eli-schwartz what happens if `-Dfoo=...` is not specified?  Is that equivalent to `-Dfoo=auto` or `-Dfoo=disabled`?  Or is that something which can be configured when the option is defined in meson_options.txt?

neheb:
feature options default to auto but can be configured in meson_options.txt. Actually can also be configured with --auto-features=X to set a default for all features.

eli-schwartz:
Like all options, you can define the default as `value: 'XXX'` in meson_options.txt

booleans default to true if unspecified, features default to auto.

gstrauss:
@neheb: based on @eli-schwartz explanation above, please modify the PR to retain the `value: false` for most options (and `value: true` for `with_pcre2` and `with_zlib`)

meson.build
```diff
@@ -1,6 +1,4 @@
-project('lighttpd', 'c', version: '1.4.68', license: 'BSD-3-Clause',
-  default_options : ['c_std=c11']
-)
+project('lighttpd', 'c', version: '1.4.68', default_options : ['c_std=c11'], meson_version: '>=0.40.0')
```
glensc: I liked the previous wrapping. but why is the license key silently removed? (not described in commit)

neheb: Rebasing issue. Will fix.

neheb: Fixed.

src/meson.build
```diff
@@ -264,146 +264,46 @@ if not(get_option('build_static'))
...
-if get_option('with_bzip')
-	libbz2 = [ compiler.find_library('bz2') ]
-	if compiler.has_function('BZ2_bzCompress', args: defs, dependencies: libbz2, prefix: '#include <bzlib.h>')
-		conf_data.set('HAVE_BZLIB_H', true)
-		conf_data.set('HAVE_LIBBZ2', true)
-	else
-		error('Couldn\'t find bz2 header / library')
-	endif
-endif
+libbz2 = dependency('bzip2', required: get_option('with_bzip'))
```
eli-schwartz, line 278: Unfortunately, this doesn't reliably work as upstream doesn't ship one. :(

The previous find_library() check is correct, here. But you can pass a feature option to find_library(..., required: ...) too.

https://sourceware.org/pipermail/bzip2-devel/2022q2/000169.html

src/meson.build
```diff
@@ -264,146 +264,46 @@ if not(get_option('build_static'))
...
 libfam = []
-if get_option('with_fam') and not(conf_data.get('HAVE_SYS_INOTIFY_H'))
-	libfam = [ compiler.find_library('fam') ]
-	if not(compiler.has_function('FAMOpen2', args: defs, dependencies: libfam, prefix: '#include <fam.h>'))
-		error('Couldn\'t find fam.h or FAMOpen2 in lib fam')
-	endif
-	conf_data.set('HAVE_FAM_H', true)
+if not conf_data.get('HAVE_SYS_INOTIFY_H')
+	libfam = [ compiler.find_library('fam', required: get_option('with_fam')) ]
+	conf_data.set('HAVE_FAM_H', libfam.found())
```
eli-schwartz, line 288: Here and later on, you do two unrelated changes:
- use feature options for simplifying found dependency/library objects
- drop the checks for header presence

You should not drop header checks, I think -- because, it's possible to have a library installed on distros with split runtime/devel packages. The runtime library will be detected by find_library, even if the development headers are not installed, and then the build fails.

eli-schwartz (changes requested):

src/meson.build
neheb: Hrm fedora does

src/meson.build
neheb: In the case of dependency, pkgconfig files are typically paired with devel packages.

neheb: hmm this might not be correct. Removing libev-devel locally but keeping libev throws this error:

```
src/meson.build:310:0: ERROR: C shared or static library 'ev' not found
```

with

```
meson --auto-features=enabled
```

eli-schwartz: Yeah, dependency() is explicitly immune to this problem.

As for removing devel packages, it really depends. Debian ships the .so symlink in the devel package, but other distros with split development packages don't.

This means that find_library() will fail when headers aren't installed... but only on Debian-like systems.

neheb: I'll test on Debian and see what's up.

neheb: I can't reproduce on Debian 11.

If I have libpam-dev, find_library succeeds. If I remove it, find_library fails. It's like that with the others as well.

gstrauss:
@neheb thank you for all this work and thank you for breaking this PR up into multiple commits.

minor: https://github.com/lighttpd/lighttpd1.4/pull/116/commits/e5e24c3d5daee911fe292a28092840f50faacd01 commit message has a double-signoff  (I can fix that)

I noticed you removed the `type: feature` for pcre2 and pcre.  Was that intentional?  My earlier comment about pcre support was only because pcre is end-of-lifed, and I did not necessarily want the new code in wrapdb downloaded, where someone might accidentally use pcre instead of pcre2.  I am fine with both pcre2 and pcre being configured using meson `type: 'feature'`, as long as pcre2 defaults to 'enabled', similar to zlib.  (I also now know a little bit more about wrapdb than I did a few days ago.)

On that note, why is https://github.com/lighttpd/lighttpd1.4/pull/116/commits/02c01c16823d6f8f2085f01f03e74aec8d5ae066 "meson: add WrapDB subprojects" part of this pull request?  I can totally see the usefulness of wrapdb, especially when building special-purpose containers, but that use case does not seem to be one that would be used by OS distros, such as openwrt, where the lighttpd build should generally obtain dependencies as shared libraries from the system/distro in order to reuse the same shared libraries, which are managed and upgraded using system/distro packaging tools.

As an alternative to that commit, would it be reasonable to simply document as a comment in the top-level meson.build that to use wrapdb, execute `meson wrap install ... <list>` and add `subproject('<dependency>')` to meson.build -- instead of having lighttpd source code ship with the subprojects/ already populated?

In general, lighttpd works with most versions of its library dependencies starting from the version of each dependency when that dependency was introduced in lighttpd.  If lighttpd does not work with the latest version of a dependency, I'd like to know about it and I'll often try to #ifdef to accomodate older and newer versions (within reason).

eli-schwartz:
> and add `subproject('<dependency>')` to meson.build

You don't need this, because e.g. `dependency('zlib')` already does everything you need:
- it will be used as a default fallback if not found on the system
- it can be force toggled with `--wrap-mode=forcefallback` (all) or `--force-fallback-for=zlib,zstd,libxml2` (specific dependencies)

Adding explicit `subproject('zlib')` would ensure configuring a subproject and making it be used, but it's not actually necessary in most cases.

gstrauss:
@eli-schwartz I think the feature is great, but the default "magic" makes me even more wary of providing a populated subprojects/ directory.  If you want subprojects/ it is easy to populate with a few simple commands which could be documented in comments in the top-level meson.build.  If one does not have shared system dependencies installed, then magically building dependencies locally might be a surprising result to many.  Later, if a package of a specific shared library is updated for a security exposure, the application-specific wrapped library build might be left exposed.

eli-schwartz:
My point was rather that you don't have to document in a comment that people edit the meson.build file to use them.

It's sufficient to run `meson wrap install ...` and they will then get picked up.

neheb:
the subprojects are part of this PR as I used them while testing that this PR works correctly. Consequently, I filed several fixes against WrapDB:

https://github.com/mesonbuild/wrapdb/commit/d515b5da81695e6c590df269173ac8aa4f4b2ef4
https://github.com/mesonbuild/wrapdb/commit/0a9eeceeee99aa232b8b01711892b6efd3d5e906

As far as pcre is concerned, how about making it a combo option with the values "auto, pcre, pcre2, disabled"?

Note that the wraps can be controlled with wrap_mode. I'm guessing you want nofallback by default? https://mesonbuild.com/FAQ.html#does-wrap-download-sources-behind-my-back

gstrauss:
@neheb
> As far as pcre is concerned, how about making it a combo option with the values "auto, pcre, pcre2, disabled"?

That sounds good.

@eli-schwartz 
> My point was rather that you don't have to document in a comment that people edit the meson.build file to use them.

Yes, ok.  I extrapolated on your comment too far.  Sorry.

@neheb 
> Note that the wraps can be controlled with wrap_mode. I'm guessing you want nofallback by default?

nofallback is the behavior in the current PR since 'fallback' is not specified in `dependency()` and `subproject()` is also not specified.

If subprojects/*.wrap are provided, then you can use `meson --wrap-mode=forcefallback` since meson 0.46.0.  That's fine and dandy and helped you find issues.  Great!

I still do not think that lighttpd needs to provide subprojects/*.wrap, as that adds (a little) bit of maintenance which I do not think is necessary because a) it is easy to `meson wrap install ...` (which can be documented in a comment in the top-level meson.build alongside some `meson configure` comments), and b) lighttpd generally aims to work with the latest releases of lighttpd's dependencies, so I do not think it too much to ask those who want to use wrapdb to `meson wrap install ...` before `meson --wrap-mode=forcefallback`

neheb:
Removed subprojects commit and made pcre a combo option.

gstrauss:
Took a quick glance and it looks good.

with_pcre2 is still in meson_options.txt, which is good.  For compatibility with existing distros using meson and `with_pcre2`, I'd like to unconditionally set pcre='pcre2'
`if pcre == 'auto' or get_option('with_pcre2')`

Would you mind if I add some comments to the top-level meson to build to `meson wrap install <list>` with the list of dependencies that you previously provided as subprojects/*.wrap files?

gstrauss:
```
--- a/meson.build
+++ b/meson.build
@@ -19,3 +19,10 @@ subdir('tests')
 
 # monolithic build (contains all plugins):
 # $ meson configure -D build_static=true -D buildtype=minsize
+
+# using https://wrapdb.mesonbuild.com/
+# $ mkdir subprojects
+# $ printf '/*/\n!packagefiles/\n' > subprojects/.gitignore
+# $ for i in liblzma libxml2 lua lz4 pcre pcre2 sqlite3 xxhash zlib zstd; do meson wrap install $i; done
+# $ for i in liblzma libxml2 lua lz4 pcre pcre2 sqlite3 xxhash zlib zstd; do meson wrap update $i; done
+# $ meson configure --wrap-mode=forcefallback ...
```

neheb:
Done.

gstrauss:
Awesome!  Thanks for working through this with me!

The patches have been committed upstream and will be pushed tomorrow (automatically closing this PR).

FYI: after merging upstream, I slightly modified the last commit to add `subprojects/` to the top-level `.gitignore`

Cheers

neheb:
np. One done so, I plan to do another PR that looks something like this:

```
--- a/src/meson.build
+++ b/src/meson.build
@@ -107,53 +107,27 @@ endif
 
 conf_data.set('HAVE_SOCKLEN_T', compiler.has_type('socklen_t', args: defs, prefix: '#include <sys/socket.h>'))
 
-if conf_data.get('HAVE_MALLOC_H')
-conf_data.set('HAVE_MALLOC_TRIM', compiler.has_function('malloc_trim', args: defs))
-conf_data.set('HAVE_MALLOPT', compiler.has_function('mallopt', args: defs))
-endif
-
-if conf_data.get('HAVE_SIGNAL_H')
-conf_data.set('HAVE_SIGACTION', compiler.has_function('sigaction', args: defs))
-conf_data.set('HAVE_SIGNAL', compiler.has_function('signal', args: defs))
-endif
-
-if conf_data.get('HAVE_SYS_EPOLL_H')
-conf_data.set('HAVE_EPOLL_CTL', compiler.has_function('epoll_ctl', args: defs))
-endif
-
-if conf_data.get('HAVE_SYS_EVENT_H')
-conf_data.set('HAVE_KQUEUE', compiler.has_function('kqueue', args: defs))
-endif
-
-if conf_data.get('HAVE_SYS_MMAN_H')
-conf_data.set('HAVE_MMAP', compiler.has_function('mmap', args: defs))
-endif
-
-if conf_data.get('HAVE_SYS_RANDOM_H')
-       conf_data.set('HAVE_GETENTROPY', compiler.has_function(
-               'getentropy',
-               args: defs,
-               prefix: '#include <sys/random.h>'
-       ))
-endif
-
-if conf_data.get('HAVE_LINUX_RANDOM_H')
-       conf_data.set('HAVE_GETRANDOM', compiler.has_function(
-               'getrandom',
-               args: defs,
-               prefix: '#include <linux/random.h>'
-       ))
-endif
-
-if conf_data.get('HAVE_SYS_RESOURCE_H')
-conf_data.set('HAVE_GETRLIMIT', compiler.has_function('getrlimit', args: defs))
-endif
-
-if conf_data.get('HAVE_SYS_UIO_H')
-conf_data.set('HAVE_PREADV', compiler.has_function('preadv', args: defs))
-conf_data.set('HAVE_PWRITEV', compiler.has_function('pwritev', args: defs))
-conf_data.set('HAVE_WRITEV', compiler.has_function('writev', args: defs))
-endif
+functions = {
+  'malloc_trim': 'malloc.h',
+  'mallopt': 'malloc.h',
+  'sigaction': 'signal.h',
+  'signal': 'signal.h',
+  'epoll_ctl': 'sys/epoll.h',
+  'kqueue': 'sys/event.h',
+  'mmap': 'sys/mman.h',
+  'getentropy': 'sys/random.h',
+  'getrandom': 'linux/random.h',
+  'getrlimit': 'sys/resource.h',
+  'preadv': 'sys/uio.h',
+  'pwritev': 'sys/uio.h',
+  'writev': 'sys/uio.h',
+}
+
+foreach f, h : functions
+  if conf_data.get('HAVE_@0@'.format(h.underscorify().to_upper()))
+    conf_data.set('HAVE_@0@'.format(f.underscorify().to_upper()), compiler.has_function(f, args: defs))
+  endif
+endforeach
 
 conf_data.set('HAVE_POLL', compiler.has_function('poll', args: defs))
 if not(conf_data.get('HAVE_POLL'))
```

Can't say I like it. It keeps behavior the same but it's probably better to change the loop to

```
+foreach f, h : functions
+  conf_data.set('HAVE_@0@'.format(f.underscorify().to_upper()), compiler.has_function(f, prefix: '#include <@0@>'.format(h), args: defs))
+endforeach
```

gstrauss:
Re: you last message above.  Sure, please file a new PR for that.  I like patterns as long as they are not too complex, or at least if the interface is easy to use, e.g. adding entries to the table.

gstrauss:
BTW, please check which version of meson added the '<string>'.format() syntax.

neheb:
https://mesonbuild.com/Reference-manual_elementary_str.html#strformat <-- I don't see version information.

I do know format() was there before the str + str syntax.

PR #117, opened 2022-09-21
[build] search for host lemon when crosscompiling with CMake

DerDakon:
Building lemon inside the project with the crosscompiler will work, but the resulting binary can't be executed on the host. Just search for a lemon binary, so the user can also override this if it is not in the path.


gstrauss:
> Building lemon inside the project with the crosscompiler will work, but the resulting binary can't be executed on the host. Just search for a lemon binary, so the user can also override this if it is not in the path.

That sounds like a different problem from the solution you provided.  lemon should be built using the native compiler, not the cross compiler.   https://wiki.lighttpd.net/InstallFromSource#cross-compiling (could use more examples from different build systems, e.g. CMake)

Have you tested lighttpd using an external lemon parser, e.g. from sqlite3 codebase?  Does it work?  I am asking because I have not tested that.

gstrauss:
https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#running-executables-built-in-the-project

DerDakon:
The problem is that CMake does not support building with 2 toolchains at the same time. When you crosscompile, you will get a target binary.

What I did was a host build of lighttpd and then point the crossbuild to use the lemon executable from that one. Well, in fact that was just an intermediate step, now I just build lemon with the host gcc by hand as it is just one file, and then point CMake to that exe.

gstrauss:
Thank you for putting this together.  It appears to work, but with action-at-a-distance that it expects lemon to be in your PATH.  I think I would prefer a slightly more explicit solution, as recommended in https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#running-executables-built-in-the-project to provide a package definition for lemon which is imported for the cross-compilation.

(Yes, building the one file by hand is also a solution, though it would be nicer for cross-compilation with CMake if a slightly better solution was made available)

gstrauss:
How about this to be more friendly cross-compiling when handling the single file that need to be build with the native compiler?
```
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -879,7 +879,17 @@ if(WIN32)
        set(COMMON_SRC ${COMMON_SRC} xgetopt.c)
 endif()
 
-add_executable(lemon lemon.c)
+if(CMAKE_CROSSCOMPILING)
+       # manually compile lemon using native compiler
+       add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/lemon
+               COMMAND cc
+               ARGS    ${CMAKE_CURRENT_SOURCE_DIR}/lemon.c -o ${CMAKE_CURRENT_BINARY_DIR}/lemon
+               DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/lemon.c
+               COMMENT "Generating lemon from lemon.c using native compiler"
+       )
+else()
+       add_executable(lemon lemon.c)
+endif()
 
 ## Build parsers by using lemon...
 lemon_parser(configparser.y)
```

gstrauss:
I am going to commit the above solution I posted, but if you have further suggestions to improve it, please post.

Thank you for helping make lighttpd better.

DerDakon:
I don't really like it because it now hardcodes "cc" instead of "lemon", but I guess it will work most of the time, so definitely better than before. Thanks.

PR #119, opened 2022-09-21
[build] let CMake handle the version number

DerDakon:
The CPack module automatically uses the version information set for the current project. While at it include the CPack module to get targets like "make package".


glensc (approved):

gstrauss:
src/CMakeLists.txt should `set(PACKAGE_VERSION "${PROJECT_VERSION}")`
(PROJECT_VERSION gets set with your proposed change to `project()`)

DerDakon:
Good catch!

PR #120, opened 2022-09-23
meson fixes 2

neheb:


neheb:
I'm curious about this clock_gettime check. has_header_symbol will never fail and so the librt logic gets skipped. I assume it should probably be something like:

```
conf_data.set('HAVE_CLOCK_GETTIME', compiler.has_function('clock_gettime', args: defs))
if not conf_data.get('HAVE_CLOCK_GETTIME')
  rt = compiler.find_library('rt')
  clock_lib = rt
  conf_data.set('HAVE_CLOCK_GETTIME', compiler.has_function('clock_gettime', args: defs, dependencies: rt))
  if not conf_data.get('HAVE_CLOCK_GETTIME')
    threads = dependency(threads')
    conf_data.set('HAVE_CLOCK_GETTIME', compiler.has_function('clock_gettime', args: defs, dependencies: threads))
    clock_lib = threads
  endif
endif
```

neheb:
https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-headers/crt/time.h#L273

I could just define _POSIX_THREAD_SAFE_FUNCTIONS directly.

gstrauss:
`_POSIX_THREAD_SAFE_FUNCTIONS` should be defined by the headers.

In the repo linked above `mingw-w64-headers/crt/unistd.h` and `mingw-w64-libraries/winpthreads/include/pthread.h` both include `pthread_unistd.h`, which sets `_POSIX_THREAD_SAFE_FUNCTIONS`.  Guessing: perhaps `#include <unistd.h>` will help detect those definitions?

gstrauss:
BTW, lighttpd already tries to detect and provide `localtime_r()` and `gmtime_r()`:
https://github.com/lighttpd/lighttpd1.4/blob/master/src/sys-time.h

On newer OS releases, localtime() and gmtime() are generally supposed to be thread-safe and reentrant.

Also, lighttpd is primarily single-threaded, so this is among the less urgent things to sort out.  (I generally prefer to code with thread-safe and reentrant routines for future-proofing.)

neheb:
Hmm? I don't see pthread_unistd.h defining _POSIX_THREAD_SAFE_FUNCTIONS. If not done in meson.build, it's not found.

Still, it's nice to have them available.

gstrauss:
_POSIX_THREAD_SAFE_FUNCTIONS is defined in the former, not the latter
./mingw-w64-libraries/winpthreads/include/pthread_unistd.h
./mingw-w64-headers/crt/pthread_unistd.h

I have not traced what determines if the former is pulled in versus the latter.

gstrauss:
As long as the header existence has already been checked before this point, why not check for header existence before trying to compile?
```
foreach f, h : functions
  if conf_data.get('HAVE_@0@'.format(h.underscorify().to_upper())
    conf_data.set('HAVE_@0@'.format(f.underscorify().to_upper()), compiler.has_function(f, prefix: '#include <@0@>'.format(h), args: defs))
  endif
endforeach
```

neheb:
has_function will fail when header is missing.

neheb:
added extra commit for muon compatibility.

gstrauss:
Sorry for taking so long to get back to this.  If you no longer have time or inclination to work on this, I'm happy to commit parts of it.  Please let me know either way with a brief note.

> I'm curious about this clock_gettime check. has_header_symbol will never fail and so the librt logic gets skipped. I assume it should probably be something like:
> 
> ```
> conf_data.set('HAVE_CLOCK_GETTIME', compiler.has_function('clock_gettime', args: defs))
> if not conf_data.get('HAVE_CLOCK_GETTIME')
>   rt = compiler.find_library('rt')
>   clock_lib = rt
>   conf_data.set('HAVE_CLOCK_GETTIME', compiler.has_function('clock_gettime', args: defs, dependencies: rt))
>   if not conf_data.get('HAVE_CLOCK_GETTIME')
>     threads = dependency(threads')
>     conf_data.set('HAVE_CLOCK_GETTIME', compiler.has_function('clock_gettime', args: defs, dependencies: threads))
>     clock_lib = threads
>   endif
> endif
> ```

I'll defer to your expertise here.  To test, I'd have to find an ancient Linux system with `clock_gettime` in `-lrt` instead of in libc (implied `-lc`).  Please go ahead and add this change to the PR.

> Why is adding '-D_POSIX_C_SOURCE' needed?

I am unable to accept that change as-is since I know there are cases where it conflicts.  Is this needed specifically for mingw?  If so let's figure out a more constrained solution for mingw, or remove from this PR.



>> As long as the header existence has already been checked before this point, why not check for header existence before trying to compile? (untested:)
>> 
>> ```
>> foreach f, h : functions
>>   if conf_data.get('HAVE_@0@'.format(h.underscorify().to_upper())
>>     conf_data.set('HAVE_@0@'.format(f.underscorify().to_upper()), compiler.has_function(f, prefix: '#include <@0@>'.format(h), args: defs))
>>   endif
>> endforeach
>> ```

> has_function will fail when header is missing.

My suggestion (theoretically) checks that the header was discovered by earlier checks, so would not run `has_function` if the required header does not exist.  If that works, then the same code can be used for `gmtime_r`, `localtime_r`, and `clock_gettime`.

Side note: please sort `functions`

Separately, I plan to remove `xgetopt.c` since it is not present in the source code and because I have provided a local win32 `getopt` implementation in lighttpd.

gstrauss:
Aha!  It appears that you added https://github.com/lighttpd/lighttpd1.4/pull/120/commits/c36b9016688b0e489ea4f10e1ad48eb1731a1502 with `-DPOSIX_C_SOURCE` for the inline functions `gmtime_r` and `localtime_r` in mingw.

...I guess that also highlights a blind spot in the other function checks.  Were those other functions to be implemented as inline functions, then such functions would not be found in the library, but rather only in the header, and so would not be detected when the function is actually present, but not as a symbol in the tested library.

gstrauss:
Come to think of it, would it be better if all the tests were `compiler.has_header_symbol()` for symbols likely to be in libc, and then `compiler.has_function()` could be used only for those symbols for which a different library might need to be specified, e.g. `-lrt` ?

gstrauss:
I just traced through headers in mingw.  lighttpd `src/sys-time.h` https://github.com/lighttpd/lighttpd1.4/blob/master/src/sys-time.h redirects `gmtime_r()` and `localtime_r()` when `_WIN32` is defined, and `_WIN32` is defined under mingw.  Therefore, https://github.com/lighttpd/lighttpd1.4/pull/120/commits/c36b9016688b0e489ea4f10e1ad48eb1731a1502 is not necessary and can be removed from this PR.

---

Some more history on `_POSIX_THREAD_SAFE_FUNCTIONS` from the mingw-w64-public mailing list:

2014:
https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/CAOYw7dtTTRrR0asBjVxVRfivqmHBmi2cAZ4TajREgRQvYxBXLA%40mail.gmail.com/

2019:
https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/fc80aa9e-012c-db28-f7a3-e424b6cd4540%40codeweavers.com/

gstrauss:
Please review the changes I pushed to the head of https://git.lighttpd.net/lighttpd/lighttpd1.4/src/branch/personal/gstrauss/master

I use `compiler.has_header_symbol()` in the loop over `functions`.

I sorted `functions` and retained `compiler.has_function()` for `clock_gettime`, removing the cygwin-specific code.

gstrauss:
FYI: I took another look through mingw headers and `_POSIX_THREAD_SAFE_FUNCTIONS` is used in a small number of places to protect `#define` of functions (e.g. `gmtime_r()`) to win32 approximations (e.g. `gmtime_s()`).  This is not needed in lighttpd, but if you wanted to extrapolate to other packages, then defining `_POSIX_THREAD_SAFE_FUNCTIONS` on mingw builds appears to be a more precise solution than defining `_POSIX_C_SOURCE`.

neheb:
Regarding the librt thing, you could just drop support for it. It seems to be an old glibc thing as far as I can tell.

Anyway, the changes LGTM.

gstrauss:
Great!  The changes on my branch will be pushed (and will auto-close this PR) in the next week or two.

BTW, when testing, I found that the openwrt/packages `net/lighttpd/Makefile` will need to be updated along with the next lighttpd release.  This is due to changes in #116 which changed some options in `meson_options.txt` from `type: 'boolean'` to `type: 'feature'`.

That change may affect other distros which use meson to build lighttpd.

gstrauss:
Thank you, @neheb!

Patches have been accepted and pushed to lighttpd master branch.

PR #121, opened 2022-09-26
clean up some more CMake constructs

DerDakon:
* remove wolfssl code that would already be handled by CMake
* use FindOpenSSL cmake module to search for OpenSSL
* use GNUInstallDirs to set defaults for several directories
* use find_package() to include the PkgConfig module
* do not check for pthread.h


gstrauss:
Have these changes been tested?
It appears that you have eliminated setting `CRYPTO_LIBRARY`, which is used later in src/CMakeLists.txt
It appears that you have eliminated setting `HAVE_OPENSSL`, which is used later in src/CMakeLists.txt
Does `find_package(OpenSSL)` do the right thing with `if(APPLE)` which defaults to using /opt/local/include?

Have these changes been tested?
There were reasons I added `include_directories(${WOLFSSL_INCLUDE_DIR} ${WOLFSSL_INCLUDE_DIR}/wolfssl)` and reasons for the other CMake logic for wolfssl.  I'd be happy to accept patches which make the CMake script cleaner, as long as things still work as expected.  Did you test building lighttpd with wolfssl using CMake after making your proposed changes?

`include(GNUInstallDirs)` is already included in the top-level CMakeLists.txt.  Does it need to be repeated in src/CMakeLists.txt?

Please do not use CaMeLcAsE unless absolutely necessary (e.g. `find_package(OpenSSL)` and `if(OpenSSL_FOUND)`)

Is `find_package(PkgConfig)` better than the more-specific `include(FindPkgConfig)`?

Removing the unused HAVE_PTHREAD_H looks fine.
Using `$CMAKE_INSTALL_*` looks good, too.

DerDakon:
> Have these changes been tested? It appears that you have eliminated setting `CRYPTO_LIBRARY`, which is used later in src/CMakeLists.txt It appears that you have eliminated setting `HAVE_OPENSSL`, which is used later in src/CMakeLists.txt Does `find_package(OpenSSL)` do the right thing with `if(APPLE)` which defaults to using /opt/local/include?

Oops, ```-D WITH_OPENSL=On``` is a bad testcase. Fixed. CMake already searches /opt/local by default, see the [Darwin platform module](https://github.com/Kitware/CMake/blob/master/Modules/Platform/Darwin.cmake#L267).

> Have these changes been tested? There were reasons I added `include_directories(${WOLFSSL_INCLUDE_DIR} ${WOLFSSL_INCLUDE_DIR}/wolfssl)` and reasons for the other CMake logic for wolfssl. I'd be happy to accept patches which make the CMake script cleaner, as long as things still work as expected. Did you test building lighttpd with wolfssl using CMake after making your proposed changes?

The code uses ```<wolfssl/wolfssl.h>``` to include these headers, so the additional path suffix should not be needed.

When you set ```-DWITH_WOLFSSL=On``` the variable can't be empty afterwards, so those cases are dead code. I just eliminated them and merged the include and library cases.

> Please do not use CaMeLcAsE unless absolutely necessary (e.g. `find_package(OpenSSL)` and `if(OpenSSL_FOUND)`)

The file is called ```FindOpenSSL.cmake```, so if I write ```find_package(OPENSSL)``` it will not be found. The ```_FOUND``` variable follows the case of the module, but for this module there is a compatibility variable with all uppercase.

> Is `find_package(PkgConfig)` better than the more-specific `include(FindPkgConfig)`?

```find_package``` is more specific, as it is about searching the things actually. It sets some more variables that can be used to control the search (which are not relevant here), but for anything normal one should use this command and not include a Find* module.

**Comment by gstrauss** [comment]
Thanks!  Many of these changes look like good improvements.

To set expectations, it will take me some time to test the proposed changes on multiple platforms and I might not get back to until next weekend.

**Comment by gstrauss** [comment]
Sorry that it has taken me so long to get back to this.  I have not forgotten, but I still need to do more testing.
To help with that, do recall on what platform(s) you tested your changes?

ISTR that /usr/local/bin was added for MacOS, building under homebrew.  Did you test on MacOS?
```
-	find_program(${_package}CONFIG_EXECUTABLE NAMES ${_package} PATHS /usr/local/bin )
+	find_program(${_package}CONFIG_EXECUTABLE NAMES ${_package})
```

**Comment by DerDakon** [comment]
> Sorry that it has taken me so long to get back to this. I have not forgotten, but I still need to do more testing. To help with that, do recall on what platform(s) you tested your changes?
> 
> ISTR that /usr/local/bin was added for MacOS, building under homebrew. Did you test on MacOS?

It was some Linux systems.

The MacOS related CMake Module [Darwin.cmake](https://github.com/Kitware/CMake/blob/master/Modules/Platform/Darwin.cmake) includes [UnixPaths.cmake](https://github.com/Kitware/CMake/blob/master/Modules/Platform/UnixPaths.cmake), which adds /usr/local to the list of known paths even in front of /usr.

File: src/CMakeLists.txt [hunk 1]
```diff
@@ -6,7 +6,7 @@ include(CheckVariableExists)
 include(CheckTypeSize)
 include(CheckLibraryExists)
 include(CMakeDetermineCCompiler)
-include(FindPkgConfig)
```
gstrauss, line 10: Should this also be removed? (not sure)
`CMakeFiles/Makefile.cmake:  "/usr/share/cmake/Modules/FindPkgConfig.cmake"`

gstrauss:
Thank you, @DerDakon 

I tested (very lightly and sparely) on a few platforms and have accepted the changes.  Thank you!

PR #122, opened 2022-10-14
[mod_proxy] use gw_plugin_config_free in mod_proxy_free_config to free gw

ggvl:
Signed-off-by: lvgenggeng <lvgenggeng@uniontech.com>

### DIFF
--- a/src/mod_proxy.c
+++ b/src/mod_proxy.c
@@ -80,6 +80,9 @@ static void mod_proxy_free_config(plugin_data * const p)
         config_plugin_value_t *cpv = p->cvlist + p->cvlist[i].v.u2[0];
         for (; -1 != cpv->k_id; ++cpv) {
             switch (cpv->k_id) {
+              case 0: /* proxy.server */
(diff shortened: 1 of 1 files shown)


gstrauss:
You changed the original title, so it appears that you were looking in this area in the code for a reason.
What problem are you trying to solve with this change?

Since lighttpd mod_proxy inherits and extends `gw_handler_ctx`, `gw_free()` must be called after everything else is handled in `mod_proxy_free_config()`, which is what the code currently does.  Your PR breaks this.  Why?

ggvl:
`cleanup` callback is called by `plugins_free()`. `gw_free()` do the same things with a for loop again. We don't want to loop twice.

gstrauss:
What problem are you trying to solve with this change?  Please be specific.

gstrauss:
You appear to have ignored my first explanation on why your patch is incorrect.

You seem to be trying to "optimize" a for loop that is expected to run on a small number of items, namely the number of scopes containing `proxy.*` configuration directives, and this loop runs at server shutdown or restart.

PR #123, opened 2022-10-23
Fix websocket HTTP/2 to HTTP/1.1 proxy

Shulyaka:
This PR fixes the websocket proxy from HTTP/2 client to HTTP/1.1 backend. Lighttpd adds a dummy Sec-WebSocket-Key value, but despite the comment, the decoded string is actually "0000000000000000\n" instead of "0000000000000000". My backend (python aiohttp library) rejects that value as being not 16 bytes long.
As the comment above the code correctly quotes, "the value of this header field MUST be a nonce consisting of a randomly selected ***16-byte value***".

### DIFF
--- a/src/mod_cgi.c
+++ b/src/mod_cgi.c
@@ -874,7 +874,7 @@ static int cgi_create_env(request_st * const r, plugin_data * const p, handler_c
 			if (!http_header_request_get(r, HTTP_HEADER_OTHER,
 			                             CONST_STR_LEN("Sec-WebSocket-Key")))
 				cgi_env_add(env, CONST_STR_LEN("HTTP_SEC_WEBSOCKET_KEY"),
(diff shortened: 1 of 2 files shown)


glensc (approved):

gstrauss:
@Shulyaka thank you.  You're correct.  There is an extra '\n' in the decoded value.

gstrauss:
@glensc, your approval means absolutely nothing to me especially when it contains zero analysis.
Feel free to add an emoji to OP's comment instead.

glensc:
@gstrauss why do you say zero analysis? I did decode both values and they match what the PR creator described.

gstrauss:
> @gstrauss why do you say zero analysis? I did decode both values and they match what the PR creator described.

I said "**contains** zero analysis".  What is posted above is "[glensc](https://github.com/glensc) approved these changes 2 hours ago" and that does not impart any useful information to me, which is what I said above.

PR #127, opened 2023-06-25
src: Fix 'getoption' meson typo

thesamesam:
Bug: https://bugs.gentoo.org/905219


gstrauss:
Thank you.  Yes, those are typos that should (and will) be fixed.

PR #128, opened 2023-06-25
mod_deflate : add support for precompressed files

alex14fr:
Add support for precompressed gzip and brotli files in mod_deflate.

This allows for very aggressive "off-line" compression of static files (e.g., using zopfli) and saves CPU cycles. This also allows serving gzip/brotli-compressed content without even compiling support for zlib/brotli in lighttpd.

### DIFF
--- a/src/mod_deflate.c
+++ b/src/mod_deflate.c
@@ -1,5 +1,21 @@
 /* mod_deflate
  *
+ * New feature in lighttpd-1.4.72 : support for precompressed files
+ *   If deflate.use_precomp is set to a positive number, then, when a
+ *   static file at path X is requested, a file named X.gz (or X.br)
+ *   is checked; if it exists, it is used as gzip (or brotli) compressed data
+ *   instead of "dynamically" compressed data (or deflate cache).
+ *   This happens even if gzip/brotli support is not compiled in, or if these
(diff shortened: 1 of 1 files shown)


gstrauss:
Your code comment includes:
```
 *   This happens even if gzip/brotli support is not compiled in, or if these
 *   encodings are not listed in deflate.allowed_encodings or the MIME type is
 *   not in deflate.mimetypes, and without consideration for system load.
```
Sounds like you do not want to use most of lighttpd mod_deflate.  lighttpd mod_deflate has an option for caching the compressed files for reuse, and reusing a cache entry includes matching the ETag of the original file, so that when the file is changed, an older (and now invalid) "precompressed" file from the cache **is not** used.  Your code does not appear to contain such precautions.

Similar to #129, your desired customizations would be better in a **custom** lua script run by [lighttpd mod_magnet](https://wiki.lighttpd.net/mod_magnet)

An advanced example of content negotiation can be found in [content-negotiation.lua](https://redmine.lighttpd.net/projects/lighttpd/wiki/AbsoLUAtion#Content-Negotiation) .  (Note: content-negotiation.lua was written before the newer lighttpd mod_magnet API in lighttpd 1.4.60 was introduced, so it should probably be updated to use the new API, but should still work with the older API.)

PR #129, opened 2023-06-27
Add support for 103 Early hints responses (RFC 8297)

alex14fr:
The aim of this module is to send the responses defined in RFC 8297 Early hints
when, for some reason, your backend is unable to send multiple responses to a single HTTP
requests (e.g., PHP-FPM <https://bugs.php.net/bug.php?id=78487>).

The headers to send are defined by the "earlyhints.headers" parameter in lighttpd configuration file.

Example:

    $HTTP["url"]=^"/slowbackend/" {
        earlyhints.headers=("Link"=>"</style.css>; rel=preload; as=style")
    }

The header value supports "file mtime substitution"; for example if you have:

    $HTTP["url"]=^"/slowbackend/" {
        earlyhints.headers=("Link"=>"</style.css?{style.css}>; rel=preload; as=style")
    }

and if <document-root>/style.css has a modification time whose Unix timestamp is 123456789, then

     Link: </style.css?123456789>; rel=preload; as=style

will be sent as Early hint header for any request with URI starting with "/slowbackend/".
This is useful if you use "cache busting".

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -91,7 +91,8 @@ builtin_mods = \
   mod_simple_vhost.c \
   mod_evhost.c \
   mod_fastcgi.c \
-  mod_scgi.c
+  mod_scgi.c \
+  mod_earlyhints.c
 
 if !LIGHTTPD_STATIC
 src += $(builtin_mods)
@@ -412,6 +413,7 @@ lighttpd_SOURCES = \
   mod_userdir.c \
   mod_vhostdb.c \
   mod_vhostdb_api.c \
+  mod_earlyhints.c \
   mod_webdav.c
 lighttpd_CPPFLAGS = \
   -DLIGHTTPD_STATIC \
--- a/src/mod_earlyhints.c
+++ b/src/mod_earlyhints.c
@@ -1,0 +1,213 @@
+#include "first.h"
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "plugin.h"
+
+#include "log.h"
+#include "buffer.h"
+#include "array.h"
+#include "request.h"
+#include "response.h"
+#include "http_header.h"
+#include "stat_cache.h"
+
+/**
+ * mod_earlyhints.c - Support for 103 Early hints responses
+ *
+ * Copyright(c) 2023 Alexandre Janon <alex14fr at gmail.com>
+ *
+ * This aim of this module is to send the responses defined in RFC 8297 Early hints
+ * when, for some reason, your backend is unable to send multiple responses to a single HTTP
+ * requests (e.g., PHP-FPM <https://bugs.php.net/bug.php?id=78487>).
+ *
+ * The headers to send are defined by the "earlyhints.headers" parameter in lighttpd configuration file.
+ *
+ * Example:
+ *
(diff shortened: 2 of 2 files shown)


gstrauss:
Thank you for your attempt at writing a lighttpd module.

There are more than a few errors, which I'll point out further below, as I try to give a code review for those who make the effort of writing code for lighttpd.

However, in short, this will not be accepted to lighttpd, as lighttpd already supports RFC 8297 and 103 Early Hints, and your customizations would be better written as a custom lua script run by [lighttpd mod_magnet](https://wiki.lighttpd.net/mod_magnet)

---

PR code comments:
* Your code includes a copyright but does not include a license.  Contributions to lighttpd should be BSD-3-clause.
* This does not appear to have been compile tested, as there are mismatched braces, resulting in invalid loops and possibly variables used outside the scope of being valid (`ds`).
* You redefinition of the global function `http_response_send_1xx()` is incorrect.  That is a global function in lighttpd's base, and not for your module to redefine.
* You do not appear to have made an attempt to match coding style of existing lighttpd modules.  Besides not using consistent spacing in your own PR, you have used tabs and added spaces between function calls and parameters.  Note: while not all lighttpd modules are themselves consistent with whitespace, most lighttpd modules do not use tabs, though some older modules have not been converted.
* Your use of `(r->conf).document_root` adds spurious parenthesis.  Since you have chosen to use lighttpd `handle_uri_clean` hook, the physical path of the request is not yet valid, and the document_root is the default inherited value, before `handle_docroot` hook has been run, and might change the document root.  This is also before the physical path has been determined, where mod_alias might have redirected the physical path and updated `r->physical.basedir` for the request.  Your modules makes the assumption that the physical path to find the .css can be determined from the inherited document root.  That's fine for your application, but does not integrate well with other lighttpd core modules.

For all these reasons, the **custom** functionality you seek should be written in a small snippet of **custom** lua code and run by [lighttpd mod_magnet](https://wiki.lighttpd.net/mod_magnet)
```
$HTTP["url"] =^ "/slowbackend/" {
    magnet.attract-raw-url-to = ( "/absolute/path/to/custom-site-early-hints.lua" )
}
```

custom-103-script.lua:
```
lighty.r.resp_header["Link"] = "</style.css>; rel=preload; as=style"
return 103
```
A more advanced example can be found on lighttpd wiki [AbsoLUAtion](https://redmine.lighttpd.net/projects/lighttpd/wiki/AbsoLUAtion) in Sample Files [early-hints.lua](https://redmine.lighttpd.net/attachments/2014)

See [lighttpd mod_magnet](https://wiki.lighttpd.net/mod_magnet) for examples how to use lighttpd lua API to `stat()` your desired path and use basic lua commands to construct a string appending your cache-busting query string.  I would recommend using the lighttpd API to append the lighttpd-generated ETag, which incorporates high-resolution timestamps (on filesystems where it is available), instead of mtime (with 1-second resolution) as you have used.  Also, your use of `snprintf()` is very expensive for string manipulation as compared to what lighttpd uses for optimized string manipulation.

gstrauss:
FYI: lighttpd supports autotools (Makefile.am), meson, CMake, and SCONS.  Additions to lighttpd generally update files for all 4 build systems, and additionally for both static and shared builds in each of the build files.

PR #130, opened 2023-12-23
[mod_openssl] use evp api for truncated sha-2 with libressl

brad0:


brad0:
I wasn't sure if I should use mod_openssl or core when looking through commits.

gstrauss:
Thanks for the PR.  I verified the EVP support in the libressl version you used.

I changed the tag to "[crypto]".  sys-crypto-md.h is not part of mod_openssl or the other modules providing TLS.  sys-crypto-md.h provides some crypto digest algorithms for the lighttpd core and some other modules, and those digest algorithms are often provided by TLS/crypto libraries such as libressl.

brad0:
> Thanks for the PR. I verified the EVP support in the libressl version you used.

Thanks.
 
> I changed the tag to "[crypto]". sys-crypto-md.h is not part of mod_openssl or the other modules providing TLS. sys-crypto-md.h provides some crypto digest algorithms for the lighttpd core and some other modules, and those digest algorithms are often provided by TLS/crypto libraries such as libressl.

Ya, I had this feeling it probably was not appropriate but all the commits I saw used the tags I listed.

PR #131, opened 2024-02-25
Fixes LEMON_PATH

birdofpreyru:
Hi @gstrauss , will you pick up this little change, please? I figured out if `CMAKE_BUILD_TYPE` is empty (happened to me when building my library for macOS (Catalyst)), your `if` condition was still entering the first branch, setting `LEMON_PATH` with double `//` in front of `lemon`, thus failing the build down the line, not able to find the executable. My change makes it work by following the second `if` branch in that case.


gstrauss:
Yes, I'll include this in the next lighttpd release -- you just missed the lighttpd 1.4.74 release by a week.

PR #135, opened 2024-06-11
[autotools] Prefer libpcre.pc to pcre-config

lgrosz:
Prefer pkg-config to configure for pcre as it seems more reliable in more situations, like cross-compilation.

---

I was running into some issues with a Yocto Project recipe, this solved that issue. I also see it had been done for the other build-systems, so figured autotools could follow suite.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -932,11 +932,16 @@ if test "$WITH_PCRE" != no; then
     PCRE_LIB="-L$WITH_PCRE/lib -lpcre"
     CPPFLAGS="$CPPFLAGS -I$WITH_PCRE/include"
   else
-    AC_PATH_PROG([PCRECONFIG], [pcre-config])
-    if test -n "$PCRECONFIG"; then
-      PCRE_LIB=`"$PCRECONFIG" --libs`
-      CPPFLAGS="$CPPFLAGS `"$PCRECONFIG" --cflags`"
-    fi
+    PKG_CHECK_MODULES([PCRE], [libpcre], [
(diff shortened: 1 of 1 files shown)


gstrauss:
Thank you.  I'll consider this patch, but please recognize that pcre is no longer supported and pcre2 should be used instead.

https://pcre.org/
> The older, but still widely deployed PCRE library, originally released in 1997, is at version 8.45. This version of PCRE is now at end of life, and is no longer being actively maintained. Version 8.45 is expected to be the final release of the older PCRE library, and new projects should use PCRE2 instead.

Yocto, yada yada, blah blah, => please go complain to Yocto to use pcre2.

lgrosz:
> Thank you.  I'll consider this patch, but please recognize that pcre is no longer supported and pcre2 should be used instead.
> 
> https://pcre.org/
> > The older, but still widely deployed PCRE library, originally released in 1997, is at version 8.45. This version of PCRE is now at end of life, and is no longer being actively maintained. Version 8.45 is expected to be the final release of the older PCRE library, and new projects should use PCRE2 instead.
> 
> Yocto, yada yada, blah blah, => please go complain to Yocto to use pcre2.

Absolutely! I must have a dependency somewhere in my system (that's already already embarrassingly out-of-date) that is preferring PCRE over PCRE2 for whatever reason. I figured I'd send the patch over just-in-case it helped one person out in a similar situation eventually. I'll work on getting my other dependencies straightened out in the meantime.

gstrauss:
The patch looks good and follows the same logic as pcre2 support.  Thanks!

PR #136, opened 2024-06-18
Small fix to use post-quantum certs with wolfSSL

anhu:


src/mod_wolfssl.c
```diff
@@ -2173,6 +2177,16 @@ network_init_ssl (server *srv, plugin_config_socket *s, plugin_data *p)
...
+#ifdef HAVE_PQC
+        if (1 != wolfSSL_CTX_use_PrivateKey_file(s->ssl_ctx,
+                                                 s->pc->ssl_pemfile->ptr,
+                                                 SSL_FILETYPE_PEM)) {
+            log_error(srv->errh, __FILE__, __LINE__,
+                      "SSL: %s %s", ERR_error_string(ERR_get_error(), NULL),
+                            s->pc->ssl_pemfile->ptr);
+            return -1;
+        }
+#else
```
gstrauss, line 2189: You should be using `s->pc->ssl_privkey->ptr` instead of `s->pc->ssl_pemfile->ptr`

You should be using `WOLFSSL_FILETYPE_PEM` instead of `SSL_FILETYPE_PEM`.  (You should prefer wolfSSL interfaces rather than the wolfSSL compatibility layer with openssl)

There is another location in the `mod_wolfssl.c` code (`mod_openssl_cert_cb()`)  which uses `wolfSSL_use_PrivateKey_buffer()` and you would want to change it to use `wolfSSL_use_PrivateKey_file()` to match your changes here.

However, none of these changes should be necessary to lighttpd mod_wolfssl.c.  Instead, you should look at the wolfSSL code for why `wolfSSL_CTX_use_PrivateKey_buffer()` is returning an error.

gstrauss:
The changes suggested in this PR should not be necessary.  Please look for bugs in wolfSSL handling cert submitted via `wolfSSL_CTX_use_PrivateKey_buffer()`

PR #137, opened 2024-06-28
[core] limit v6only warning for addresses like ::1

Arusekk:
Addresses starting with two colons are identifying v6 addresses, just as if there were brackets around them.  The warning is not relevant in that case.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -414,7 +414,7 @@ static int network_server_init(server *srv, const network_socket_config *s, buff
 	}
 
 	host = host_token->ptr;
-	if ((use_ipv6 && (*host == '\0' || *host == ':')) || (host[0] == '[' && host[1] == ']')) {
+	if ((use_ipv6 && (*host == '\0' || (host[0] == ':' && host[1] != ':'))) || (host[0] == '[' && host[1] == ']')) {
 		log_warn(srv->errh, __FILE__, __LINE__,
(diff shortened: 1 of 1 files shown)


stbuehler:
> Addresses starting with two colons are identifying v6 addresses, just as if there were brackets around them.

I don't think so. Afaict we require address + port in this place, so IPv6 addresses always need brackets. The "shouldn't start with `:`"-check tests for an empty address. `::1` should get parsed as host `:` and port `1`; that shouldn't work out.

Arusekk:
I've actually used the following config in production for several years, and it worked as expected (binding on `[::1]:8080`).
```
server.port = 8080
server.bind = "::1"
```

I had to change to `server.bind = "[::1]"` though to avoid the warning.

gstrauss:
https://wiki.lighttpd.net/Server_use-ipv6Details

Please read the doc and consider removing `server.use-ipv6` from your lighttpd.conf.  Add `server.v4mapped` if needed.

---

> I've actually used the following config in production for several years, and it worked as expected (binding on `[::1]:8080`).
> 
> ```
> server.port = 8080
> server.bind = "::1"
> ```
> 
> I had to change to `server.bind = "[::1]"` though to avoid the warning.

In your patch, you failed to mention what version of lighttpd you were using and now what version of lighttpd you are using.

PR #138, opened 2024-07-12
[h2] fix spurious connection resets with zero log_monotonic_secs

MartinNowak:
- apparently monotonic clock starts at zero on Cloud Run's gVisor
  - potentially affects other hypervisors/container-emulations
- h2 sent_settings protocol flag was used both as bool and timestamp which lead to a protocol error[¹](https://git.lighttpd.net/lighttpd/lighttpd1.4/src/a578c50d78a3085fe78d0ed97fb3e42affe9936a/src/h2.c#L1047-L1048) when receiving a settings ACK while log_monotonic_secs was still zero
- fixed by using at least 1s as timestamp for sent_settings


gstrauss:
It seems that it is possible to boot a cloud system in < 1 second.

Thank you for the patch.

PR #139, opened 2024-07-19
[mod_auth] Fix nonce validation with nonce_secret

ToMe25:
In my testing I couldn't get digest authentication with `nonce-secret` set working from lighttpd 1.4.69 all the way to 8ad583c, though admittedly I didn't test any versions in between.

If the credentials were correct, it would fail with a 401 response and a log message like this:
```
2024-07-19 16:00:02: (mod_auth.c.1488) digest: nonce invalid
```

After some debugging I found a total of three issues which seem to have caused this.  
These issues and their solutions are described below the MRE.

Here is a minimal config I used to reproduce the problem:
```
server.modules += ( "mod_auth", "mod_authn_file" )

server.document-root = "/path/to/project/root/"
server.port = 8080

mimetype.assign = (
                    ".txt"  => "text/plain; charset=utf-8",
                    ".html" => "text/html; charset=utf-8",
                    ".json" => "application/json; charset=utf-8",
)

# Auth config
auth.backend = "plain"
auth.backend.plain.userfile = "/path/to/userfile"
auth.require = ( "/api/" =>
                 ( "method"	  => "digest",
                   "realm"        => "example realm",
                   "require"	  => "valid-user",
                   "nonce-secret" => "Very Secure Secret",
                 )
)
```

In addition to this config a userfile is required.

Using this config all attempts at authenticating a request to `/api/` failed, while removing the `nonce-secret` caused them to work perfectly.

## Issues

The first issue I found was that the nonce that is generated for validation is generated with a new timestamp, rather than the parsed timestamp.  
This means that automated requests that are fast enough that the timestamp doesn't change could succeed, but almost all requests would fail.

I fixed this by using `ts` instead of `cur_ts` when generating the nonce, which shouldn't cause any issues as far as I can tell but resolves this issue.
<br>

The next issue is that the colon between the timestamp and the random data wasn't skipped during parsing, causing the parsed random data to always be zero.  
I fixed this be adding a `i++` before the parsing of the random data.  
I am pretty certain that this too doesn't cause any issues.

This incrementing of the value could of course be moved somewhere else, but I wasn't able to figure out where it should be to best match the code style.

With these two fixed approximately 50% of requests succeeded.
<br>

The third and final change I implemented was making the `rng` variable in `mod_auth_append_nonce` unsigned.  
Without this change the cast to  `uintmax_t` meant that the negative values would be prefixed with `ffffffff`, which in turn prevented them from being parsed to an unsigned integer.  
I am uncertain whether this change causes conversion issues with the case without `nonce-secret`(because `li_rand_pseudo` still returns an int), but since the nonce doesn't get validated in that case those shouldn't matter.

I validated that authentication without `nonce-secret` still works with these changes.


ToMe25:
Should I add a cast to `unsigned int` to the `li_rand_pseudo` call?  
I didn't add one because my Compiler didn't complain, but I just remembered that this is theoretically an implicit cast, which some projects don't like.

PS. Should I have created an issue about this before making this PR?  
I couldn't find anything about whether the lighttpd project prefers that.

gstrauss:
Thank you for the very thorough write-up, including test case, patches, and explanation!

I don't see how the original code ever could have worked with nonce_secret set to a non-empty string due to `i` not being incremented past the `':'` separator, so not included in the digest used for validation.

> This incrementing of the value could of course be moved somewhere else, but I wasn't able to figure out where it should be to best match the code style.

That place is fine.  I would prefer to put it further above during the check for `':'` to indicate that the `':'` had been checked and the next position is ready, whether nonce_secret is set or (theoretically) something else parses the next char.

> The third and final change I implemented was making the rng variable in mod_auth_append_nonce unsigned.
Without this change the cast to uintmax_t meant that the negative values would be prefixed with ffffffff, which in turn prevented them from being parsed to an unsigned integer.

You're correct.  This would work, too:
```
-        buffer_append_uint_hex(b, (uintmax_t)rnd);
+        buffer_append_uint_hex(b, (uintmax_t)(unsigned int)rnd);
```

However, I think it is cleaner to use your `unsigned int rnd;` and to change the signature of the function to take `unsigned int *rndptr`, and finally to cast 
```
-        rnd = rndptr ? *rndptr : li_rand_pseudo();
+        rnd = rndptr ? *rndptr : (unsigned int)li_rand_pseudo();
```

Here's my changes to your patch, at the moment untested:
```
--- a/src/mod_auth.c
+++ b/src/mod_auth.c
@@ -988,14 +988,14 @@ mod_auth_digest_mutate (http_auth_info_t * const ai, const http_auth_digest_para
 
 
 static void
-mod_auth_append_nonce (buffer *b, unix_time64_t cur_ts, const struct http_auth_require_t *require, int dalgo, int *rndptr)
+mod_auth_append_nonce (buffer *b, unix_time64_t cur_ts, const struct http_auth_require_t *require, int dalgo, unsigned int *rndptr)
 {
     buffer_append_uint_hex(b, (uintmax_t)cur_ts);
     buffer_append_char(b, ':');
     const buffer * const nonce_secret = require->nonce_secret;
-    int rnd;
+    unsigned int rnd;
     if (NULL == nonce_secret)
-        rnd = rndptr ? *rndptr : li_rand_pseudo();
+        rnd = rndptr ? *rndptr : (unsigned int)li_rand_pseudo();
     else { /*(do not directly expose random number generator single value)*/
         rndptr
           ? (void)(rnd = *rndptr)
@@ -1454,7 +1454,7 @@ mod_auth_digest_validate_nonce (request_st * const r, const struct http_auth_req
      * that is highly unlikely within a 10 min (moving) window of valid
      * time relative to current time (now).
      * If it is desired to validate that nonces were generated by server,
-     * then specify auth.require = ( ... => ( "secret" => "..." ) )
+     * then specify auth.require = ( ... => ( "nonce-secret" => "..." ) )
      * When secret is specified, then instead of nanoseconds, the random
      * data value (included for unique nonces) will be exposed in the nonce
      * along with the timestamp, and the additional secret will be used to
@@ -1466,7 +1466,7 @@ mod_auth_digest_validate_nonce (request_st * const r, const struct http_auth_req
         ts = (unix_time64_t)((uint64_t)ts << 4) | hex2int(nonce[i]);
 
     const unix_time64_t cur_ts = log_epoch_secs;
-    if (nonce[i] != ':' || ts > cur_ts || cur_ts - ts > 600) { /*(10 mins)*/
+    if (nonce[i++] != ':' || ts > cur_ts || cur_ts - ts > 600) { /*(10 mins)*/
         /* nonce is stale; have client regenerate digest */
         return mod_auth_send_401_unauthorized_digest(r, require, ai->dalgo);
     }
@@ -1488,7 +1488,7 @@ mod_auth_digest_validate_nonce (request_st * const r, const struct http_auth_req
         }
         buffer * const tb = r->tmp_buf;
         buffer_clear(tb);
-        mod_auth_append_nonce(tb, cur_ts, require, ai->dalgo, (int *)&rnd);
+        mod_auth_append_nonce(tb, ts, require, ai->dalgo, &rnd);
         if (!buffer_eq_slen(tb, dp->ptr[e_nonce], dp->len[e_nonce])) {
             /* nonce not generated using current require->nonce_secret */
             log_error(r->conf.errh, __FILE__, __LINE__,
```

ToMe25:
I added your changes as a new commit authored by you.

I'm not sure if that is the right way to commit them, so feel free to tell me a different way to do so :)

I also tested it again, and it works with and without `nonce-secret`.

gstrauss:
> I'm not sure if that is the right way to commit them, so feel free to tell me a different way to do so :)

You're fine.  Thanks for asking.  I'd like to give you credit.  My process is to squash the commits into your commit, and will add a Co-authored line to the commit, as well as have the commit automatically close this PR.

gstrauss:
>  PS. Should I have created an issue about this before making this PR?
> I couldn't find anything about whether the lighttpd project prefers that.

If you have a patch, one PR is fine.
The context, reproduction case, and explanation you provided with this patch is excellent.

PR #141, opened 2025-02-16
[mod_dirlisting] Fixes for mtime sorting and query parameter sort order

Phaze-III:
### Issues
Sorting on mtime in lighttpd-1.4.77 somehow sorts on day of month without using the month. It appears that the year is used to lookup the month-index. Also using the &O=A did a descending sort and &O=D ascending. This PR fixes both issues.

Patches tested with lighttpd-1.4.77 on FreeBSD 14.2 with Chrome and Firefox (MacOS and Android) and Safari (Mac OS)

### DIFF
--- a/src/mod_dirlisting.c
+++ b/src/mod_dirlisting.c
@@ -756,12 +756,12 @@ static const char js_simple_table_resort[] = \
 "    el.dataset.value = -1;\n" \
 "  } else {\n" \
 "   var d = li_date_regex.exec(get_inner_text(el));\n" \
-"   el.dataset.value = (parseInt(d[0])*(1<<25))\n" \
-"     + ( (li_mon_num(d[1])<<22)\n" \
-"        |(parseInt(d[2])  <<17)\n" \
-"        |(parseInt(d[3])  <<12)\n" \
-"        |(parseInt(d[4])  << 6)\n" \
-"        |(parseInt(d[5])) );\n" \
+"   el.dataset.value = (parseInt(d[1])*(1<<25))\n" \
+"     + ( (li_mon_num(d[2])<<22)\n" \
+"        |(parseInt(d[3])  <<17)\n" \
+"        |(parseInt(d[4])  <<12)\n" \
+"        |(parseInt(d[5])  << 6)\n" \
+"        |(parseInt(d[6])) );\n" \
 "  }\n" \
 " }\n" \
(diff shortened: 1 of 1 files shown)


gstrauss:
Thank you.  Sorry for the delayed response.  I'll make time to review before the end of the coming week.

Phaze-III:
Thanks for looking into this. To illustrate the issues I made a few screenshots of a test-directory on an unpatched server with files generated once a day with a <YYYY-MM-DD_hh:mm:ss>.html formatted filename for a few months.

### mtime sorting issue
Default listing, sorted on filename ascending:

Screenshot 2025-02-22 at 16 47 37

Listing after clicking on the `Last modified:` column, should result in sorted ascending on mtime and effectively the same list as in the previous screenshot. Sorting is however done on a combination of year, day-of-month and time, without using the month:

Screenshot 2025-02-22 at 16 48 09

Same for descending sort on mtime, first few entries look ok but that is because there are no files yet for Feb 23 to Feb 28:

Screenshot 2025-02-22 at 16 48 40

### O=[A|D] sort order issue

Given that the default sort order on filename is ascending using `?C=N&O=A` should give a list ordered from oldest to newest (using the ISO-style filename). However it is from newest to oldest (descending):

Screenshot 2025-02-22 at 16 50 19

Vice versa for `?C=N&O=D` should be descending (newest to oldest) but is ascending (oldest/lowest to newest/highest):

Screenshot 2025-02-22 at 16 50 46

gstrauss:
Patch looks good.  Thank you!

The original use of javascript regex captures (which you fixed) was off-by-one.  The `d[0]` index is the match, and the captures start at `d[1]`.

Also, as you note, the result of ascending and descending when requested by ?C=N&O=A or O=D was reversed.

There are mixed opinions on the internet whether the arrow next to the column name should indicate up or down for ascending or descending, and whether or not it should indicate the current state or what happens when it gets clicked.

I am going to leave it as it currently is, even though the patch fixes the results for ascending and descending.  If anyone with a UX background has a strong opinion, please file another PR with a patch and explanation.

https://ux.stackexchange.com/questions/37564/use-up-or-down-arrow-to-represent-sort-ascending-at-table-header
https://stackoverflow.com/questions/338166/which-direction-should-the-arrows-point-in-a-sorted-table

PR #142, opened 2025-04-16
workaround unsupported PR_CAP_AMBIENT_CLEAR_ALL on Cloud Run V1

MartinNowak:
- would similarly fail on linux kernels < 4.3
- unfortunately errno is not very specific
  > EINVAL The  value of op is not recognized, or not supported on this system.
  > EINVAL An unused argument is nonzero.

Happy to maintain that patch myself, but thought I'd try to share it.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1901,7 +1901,9 @@ static int server_main_setup (server * const srv, int argc, char **argv) {
 #if defined(HAVE_SYS_PRCTL_H) && defined(PR_CAP_AMBIENT)
 	/* clear Linux ambient capabilities, if any had been granted
 	 * (avoid leaking privileges to CGI or other subprocesses) */
-	if (prctl(PR_CAP_AMBIENT, PR_CAP_AMBIENT_CLEAR_ALL, 0L, 0L, 0L) < 0) {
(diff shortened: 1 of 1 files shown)


gstrauss:
BTW, `defined(PR_CAP_AMBIENT)` check should catch lighttpd compiled against Linux kernel < 4.3, so your patch is more aimed at emulators which do not support this `prctl` (and at compilations against mismatched kernel headers and kernel, which is its own bad idea).

I may be more paranoid and try `PR_CAP_AMBIENT_IS_SET` if `PR_CAP_AMBIENT_CLEAR_ALL` fails.  If `PR_CAP_AMBIENT_IS_SET` succeeds (0 or 1 instead of -1 and EINVAL), then still error out, since dropping capabilities was not supported.  Would you be willing to test this on Cloud Run v1?

gstrauss:
Eh, I have reconsidered and will accept your patch as-is.  The args to `prctl` are supported or not.  If `EINVAL`, then it would be reasonable to assume Linux ambient capabilities are not supported by the kernel.

gstrauss:
Thank you, @MartinNowak

PR #144, opened 2025-08-14
Unifying chunk buffer size logic

4oomin:
Hi, I'm an engineer who use your lighttpd1.4
It would not be a bug on your side. But based on my experience, I think it can cause critical issue.

# [1] Issue in the original code
1
Oversized buffer size logic is different between **chunk_buffer_acquire_sz()** and **chunk_acquire().**

2
According to **chunk_acquire()**
Every number of sz become size up to the nearest multiple number of **chunk_buf_sz.**
But only when sz is 8193, it size down to the nearest multiple number of **chunk_buf_sz** 8192.

# [2] What I fix
I unify all the rule of sizing chunk buffer as sizing up to the nearest multiple number of **chunk_buf_sz.**
I know you try to avoid excess allocation by sizing down sz when sz is **chunk_buf_sz*N+1**
I explain below why I fixed it this way.

# [3] Why I do this
1
Until recently, I use lighttpd 1.4.52 which is 7 years ago version.
With this version when ethernet speed is over 1Gbps, lighttpd fail to fork because of lack of memory.
I found out **chunkqueue_append_buffer_open_sz()** in chunk.c was related with it
It allocated new memory whenever sz is 4KB and sz is always 4KB when using 1Gbps ethernet.
So it caused memory leak.
I know it's not directly related with this issue, but these two issue are related with allocating additional space for '\0'
So it would be better to give additional space for '\0' regardless of the size of sz.

2
As I understand, it doesn't give extra space for '\0' when sz is **chunk_buf_sz*N+1**. 
But it allocate extra space if sz is not **chunk_buf_sz*N+1**. So I think it lose some consistency.
 
Thank you

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -352,7 +352,7 @@ __attribute_returns_nonnull__
 static buffer * chunk_buffer_acquire_sz(const size_t sz) {
     chunk *c;
     buffer *b;
-    if (sz <= (chunk_buf_sz|1)) {
+    if (sz <= chunk_buf_sz) {
         if (chunks) {
             c = chunks;
             chunks = c->next;
@@ -368,7 +368,7 @@ static buffer * chunk_buffer_acquire_sz(const size_t sz) {
              * (sz & ~1uL) relies on buffer_realloc() adding +1 *and* on callers
              * of this func never passing power-2 + 1 sz unless direct caller
              * adds +1 for '\0', as is done in chunk_buffer_prepare_append() */
-            c = chunk_init_sz(((sz&~1uL)+(chunk_buf_sz-1)) & ~(chunk_buf_sz-1));
+            c = chunk_init_sz((sz + (chunk_buf_sz-1)) & ~(chunk_buf_sz-1));
         }
     }
     c->next = chunk_buffers;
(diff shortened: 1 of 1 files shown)


gstrauss:
> As I understand, it doesn't give extra space for '\0' when sz is chunk_buf_sz*N+1.
But it allocate extra space if sz is not chunk_buf_sz*N+1. So I think it lose some consistency.

I have had a very hard time following your writing and your justifications.  I do not understand your suggestions.

> I know this change is for avoiding excess allocation like avoiding to allocate 24KB when 16K+1B is required.

24KB is not a power-of-2, so I do not understand how you could have arrived at that statement.

---

`chunk_buffer_acquire_sz()` is a static function in `chunk.c` and the callers in `chunk.c` are either requesting `chunk_buf_sz` or a sz which includes +1 for `'\0'`. 

`chunk_acquire()` is a static function in `chunk.c`, though the sz passed to it may be passed into public functions in `chunk.c` which then call `chunk_acquire()`.  Those public functions are `chunkqueue_prepend_buffer_open_sz()`, `chunkqueue_append_buffer_open_sz` and `chunkqueue_get_memory()`.  However, most callers already include the +1 or the +1 does not matter since the size requested is an estimate, e.g. 1k or 4k.

Remaining calls of interest to me:

I might take a closer look at `h2_send_cqdata()` call to `chunkqueue_append_buffer_open_sz()`.

=> `chunkqueue_get_memory()` appears as if it might benefit from adding +1 to sz when requesting allocation via `chunkqueue_append_buffer_open_sz()`, and this suggestion is much, much, much, much, much simpler than the tome you wrote.

PR #151, opened 2026-02-02
fix for PCRE2 find_package()

pkleymonov-qnx:
Tries to find pcre2 with help of cmake native method:  find_package()
it helps for cross-compilation and useful with conan package manger


gstrauss:
I have not had time to look into expanding this into other parts of CMakeLists.txt, but that does not prevent me from accepting this patch.  It will be part of the next lighttpd release.  Thank you.

PR #152, opened 2026-03-11
openssl: start using OpenSSL3 providers for certificate and key

amananas:
OpenSSL v3 introduced store providers as the go-to way to retrieve certificate(s) and keys from any source.
The main point of this change is, abstracting the storage-specific code for data retrieval in an OpenSSL provider, which is not known from the application itself (instead of OpenSSL engines which require per engine specific code).

This change leverages this feature, using it for all key and certificates loading if OpenSSL version is compatible. Default (file) storage and custom (HSM, ...) key storage should work transparently, assuming OpenSSL configuration loads the needed providers.
The implementation itself is based on OpenSSL applications source code as reference.

No change is required from the user to keep using certificates and keys from files, since the default provider available in OpenSSL supports file URI (and assume a file URI if no scheme is specified: `/some/file.pem` -> `file:/some/file.pem`).

### DIFF
--- a/src/mod_openssl.c
+++ b/src/mod_openssl.c
@@ -70,6 +70,10 @@
 #ifndef OPENSSL_NO_OCSP
 #include <openssl/ocsp.h>
 #endif
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#include <openssl/provider.h>
+#include <openssl/ui.h>
+#endif
 #ifdef AWSLC_API_VERSION /* alt: OPENSSL_IS_AWSLC */
 /* AWS-LC derived from BoringSSL, but AWSLC_API_VERSION has different meaning.
  * Reuse BORINGSSL_API_VERSION for (presently) small num of API version checks*/
@@ -2355,7 +2359,7 @@ mod_openssl_cert_is_active (const X509 *crt)
   #endif
 }
 
-
+#if OPENSSL_VERSION_NUMBER < 0x30000000L || defined(TLSEXT_TYPE_application_layer_protocol_negotiation)
 static X509 *
 mod_openssl_load_pem_file (const char *file, log_error_st *errh, STACK_OF(X509) **chain)
 {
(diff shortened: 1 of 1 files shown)


gstrauss:
Apologies that I have not come back to this for so long.

ISTR conversations elsewhere, so please forgive me as I ask some questions again to document things here:

What is the provenance of the code in this patch?  Is it copied from anywhere?  The coding style suggests to me that it is copied from openssl code.

Why modify `network_openssl_load_pemfile()` instead of having `mod_openssl_evp_pkey_load_pem_file()` and `mod_openssl_load_pem_file()` call into the new routine `mod_openssl_load_store_data()` with appropriate arguments?  The original routines are used for loading site certificates as well as for acme-tls/1.  Is that the reason you modified the calls only in `network_openssl_load_pemfile()`?

One of the reasons that lighttpd loads the on-disk files is because, for portability, openssl uses stdio routines, and portable stdio routines use caching.  lighttpd bypasses the stdio caching layer and loads potentially sensitive files into memory, and then wipes that memory after the files are processed.

Do the `OSSL_STORE_*` functions handling caching of sensitive files data properly?  (I have not looked in the openssl code.)  Would it be a better transition to check the uri/name to see if it starts with a scheme, and to use `mod_openssl_load_store_data()` if the uri begins with a recognized scheme?

gstrauss:
I asked above:
> What is the provenance of the code in this patch? Is it copied from anywhere? The coding style suggests to me that it is copied from openssl code.

The commit message in this PR contains:
```
The implementation itself is based on OpenSSL applications source code
as reference and should be updated upon changes.
```

It appears to me that the code is copied and based off of openssl source code `apps/lib/apps.c:load_key_certs_crls()`, but this PR appears to introduce a few mistakes.  In the error path, value inside the ptr-ptr is not reset to NULL after any of the openssl `*_free(*ptr)` functions are called.  More importantly, the storing of the certificate and building of the certificate chain has an issue.  The primary cert is duplicated into the certificate chain stack.

PR #153, opened 2026-03-23
Cap snprintf return value in mod_maxminddb to prevent stack over-read

kodareef5:
snprintf at lines 379/382 returns the number of characters that would have been written, which can exceed sizeof(buf) for large double/float values. This value is used directly as the data length for http_header_env_set, causing a stack buffer over-read. Cap vlen to sizeof(buf) - 1 after each snprintf call.


gstrauss:
Did you find any double/float values in the maxminddb GeoIP2 database that exceed the buffer size in mod_maxminddb.c (`buf[34]`) when using `snprintf` using format string `"%.5f"` to stringify the double/float value?

PR #155, opened 2026-04-28
Reject odd-length hex to prevent reading data from outside of buffer.

elnerd:
When the input hex has odd length, li_hex2bin will read outside of the input buffer.
Added check to disallow odd-length buffers.


gstrauss:
Current usage of `li_hex2bin()` inside lighttpd auth modules will not pass an odd length.

In any case, the sanity check is reasonable and I have accepted your patch.  Thank you.

In the future, please match the whitespace style in with the surrounding code.  Your one-line patch changed 4 spaces to a tab, and had extraneous trailing whitespace at the end of the line.

PR #157, opened 2026-05-24
[mod_dirlisting] faster client-side sort on huge listings

cicku:
resort() ends with:

    for (var i=0;i<rows.length;i++)
        table.tBodies[0].appendChild(rows[i]);

appendChild() on a row already inside tbody scales with the number of siblings, so re-attaching N rows is O(N^2).  For a 29k-entry listing Chromium hangs for ~44s (Example: https://mirrors.cicku.me/archlinux/extra/os/x86_64/).  Detaching tbody first does not help; neither does replaceChildren().

Clear tbody in one mutation (innerHTML = ""), then re-attach the sorted rows through a single DocumentFragment.  Same listing drops to ~100ms.  No measurable regression at small N.

While here:

  - emit data-value on the date cell (mtime as epoch) and the type cell (0 for directory, 1 for file).  size already had one.  The comparator now uses pre-computed keys for every column except Name, and li_dates_to_dv() with its date regex / month table is no longer needed.

  - get_inner_text() returns textContent rather than innerText, so a read does not force layout.  dataset.value above covers everything other than the Name column.

  - init_sort_from_query() defaults to column = -1 when ?C= is not set.  init_sort() then just turns the headers into sort links without touching the row order; the server has already emitted rows in the configured order (alphabetical, or readdir() under dir-listing.sort = "disable").  Existing ?C=N|M|S|T|D and ?O=A|D URLs are unaffected and still display the correct arrow.

### DIFF
--- a/src/mod_dirlisting.c
+++ b/src/mod_dirlisting.c
@@ -749,61 +749,19 @@ static const char js_simple_table_resort[] = \
 "    && (typeof el.dataset.value === 'string'\n" \
 "        || typeof el.dataset.value === 'number'))\n" \
 "  return el.dataset.value;\n" \
-" if(el.innerText)\n" \
-"  return el.innerText;\n" \
-" else {\n" \
-"  var str = \"\";\n" \
-"  var cs = el.childNodes;\n" \
-"  var l = cs.length;\n" \
-"  for (var i=0;i<l;i++) {\n" \
-"   if (cs[i].nodeType==1) str += get_inner_text(cs[i]);\n" \
-"   else if (cs[i].nodeType==3) str += cs[i].nodeValue;\n" \
-"  }\n" \
-" }\n" \
-" return str;\n" \
-"}\n" \
-"\n" \
-"var li_date_regex=/(\\d{4})-(\\w{3})-(\\d{2}) (\\d{2}):(\\d{2}):(\\d{2})/;\n" \
-"\n" \
-"var li_mon = ['Jan','Feb','Mar','Apr','May','Jun',\n" \
-"              'Jul','Aug','Sep','Oct','Nov','Dec'];\n" \
-"\n" \
-"function li_mon_num(mon) {\n" \
-" var i; for (i = 0; i < 12 && mon != li_mon[i]; ++i); return i;\n" \
-"}\n" \
-"\n" \
-"function li_dates_to_dv(table) {\n" \
-" for (var j=1;j<table.rows.length;j++) {\n" \
-"  var el = table.rows[j].cells[date_column];\n" \
-"  if(el.dataset && typeof el.dataset.value != 'undefined') {\n" \
-"   if (typeof el.dataset.value == 'number')\n" \
-"    break;\n" \
-"   if (el.dataset.value === \"-1\")\n" \
-"    el.dataset.value = -1;\n" \
-"  } else {\n" \
-"   var d = li_date_regex.exec(get_inner_text(el));\n" \
-"   el.dataset.value = (parseInt(d[1])*(1<<26))\n" \
-"     + ( (li_mon_num(d[2])<<22)\n" \
-"        |(parseInt(d[3])  <<17)\n" \
-"        |(parseInt(d[4])  <<12)\n" \
-"        |(parseInt(d[5])  << 6)\n" \
-"        |(parseInt(d[6])) );\n" \
-"  }\n" \
-" }\n" \
+" return el.textContent;\n" \
 "}\n" \
 "\n" \
 "function sortfn_then_by_name(a,b,sort_column) {\n" \
 " if (sort_column == name_column || sort_column == type_column) {\n" \
-"  var ad = (a.cells[type_column].innerHTML === 'Directory');\n" \
-"  var bd = (b.cells[type_column].innerHTML === 'Directory');\n" \
-"  if (ad != bd) return (ad ? -1 : 1);\n" \
+"  var ad = a.cells[type_column].dataset.value;\n" \
+"  var bd = b.cells[type_column].dataset.value;\n" \
+"  if (ad !== bd) return (ad < bd ? -1 : 1);\n" \
 " }\n" \
 " var at = get_inner_text(a.cells[sort_column]);\n" \
 " var bt = get_inner_text(b.cells[sort_column]);\n" \
 " var cmp;\n" \
-" if (sort_column == size_column) {\n" \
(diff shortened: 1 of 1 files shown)


cicku:
Patched version is live at: http://demo.cicku.me/lighty/

gstrauss:
Impressive performance!  Thank you!

I will experiment with this.  Adding an additional td data-value per row does slightly increase the size of the page, and creating the document fragment during sorting might duplicate the table an additional time, but the increase in transferred page size and the increase in memory use during sorting on the client are likely worth the tradeoff for less CPU during sorting on the client (due to elimination of per-row reflows) and much faster client performance for sort.

gstrauss:
Initial testing with Firefox on Linux on a 10-year-old laptop connecting to lighttpd running on localhost:

Page size for 100,000 files increases over 22% from over 17M to over 21M and this affects initial page load time.

Sort time (again, on Firefox) appears approx 10% faster (not precisely measured), but not the order of magnitude you reported on Chrome.

For 3400 files, the number of files in my /usr/bin, I did not notice the difference in sort performance, which is why I increased to 100000.  FYI: I created my test directory using:
`mkdir t; cd t; for i in (seq 1 100000); do echo $i; done | xargs -P 4 touch`

I need to test on Chrome next.  (AFK; I'll come back to this soon.)

gstrauss:
Wow!  On a 14-year-old laptop running Windows, Chrome is quite the dog on a large dirlisting page.  I had to reduce the test to 30000 files, and yes, your PR is much faster to load the initial page, as well as to sort, though only *after* the first sort.  The first sort using Chrome with your PR takes some 23 seconds on this laptop, and subsequent sorts take 2-3 seconds.

By comparison with Firefox on Windows on this laptop, the 30000 file test performed about the same with old and new javascript, taking 2-3 seconds for sort.  The new javascript seems slightly faster.  Now, with Firefox and the 100000 file test, the new javascript took 7-8 seconds for sort, whereas the old took significantly longer.

The above is mostly quick tests with coarse timing precision.  Clearly, avoiding the reflows makes sense.

To reduce page size, I might experiment with avoiding the data-value for mtime, or calculating it after the page has been downloaded.  (The data-value for size is still needed because otherwise the exact file sizes are lost in the display summary which rounds to K, M, G, etc.)

gstrauss:
To reduce the size of the page, I'd like to remove from the patch the additions of data-value you made for mtime and type.  The data-value can be calculated for mtime upon the first sort of that column, *if it ever occurs*.

I still want to test the data-value on the type column, but if keeping the rest of the patch requires the data-value, then that, too, can be generated once, upon first use.

---

Are you able to reproduce the behavior I am seeing with Chrome (on my old, slow laptop) where the *first* sort (of any column) takes 23-28 seconds, but subsequent sorts take 2-3 seconds?

gstrauss:
I modified your PR and added a commit to a dev branch at
https://github.com/gstrauss/lighttpd1.4/tree/mod_dirlisting

For common use which does not use sorting, omitting the data-value on mtime and type columns measurably reduces response size, so I have removed those from the patch.

The first triggered sort on Chrome is slooooooow.  For this reason, I am considering re-enabling the prior behavior of triggering a sort as part of the initial javascript execution on the page.  Do you have any hint why Chrome is so slow on the first triggered sort (any column), but then 2-3 seconds for subsequent sorts?  It is slow even on the size column, which is not special-cased and has data-value set on the server-side and delivered with the page.

cicku:
Thank you for the modified commit!

I have the same struggle - I could not notice the hanging on Mobile Safari, Firefox or Opera. This issue seems to be a bit specific to Chrome, perhaps I need to grep through their issue tracker - or even submit a report there.

My local test is from a M4 macbook, and do not have access to old laptops anymore.

gstrauss:
BTW, when I backported the following from lighttpd2 in 2016, ISTR that the parseInt() was necessary for proper functionality.  It seems that is no longer needed since my 2024 commit which introduced data-value for size.
https://redmine.lighttpd.net/issues/3264
```
@@ -801,9 +790,7 @@ static const char js_simple_table_resort[] = \
 " var at = get_inner_text(a.cells[sort_column]);\n" \
 " var bt = get_inner_text(b.cells[sort_column]);\n" \
 " var cmp;\n" \
-" if (sort_column == size_column) {\n" \
-"  cmp = parseInt(at)-parseInt(bt);\n" \
-" } else if (sort_column == date_column) {\n" \
+" if (sort_column == size_column || sort_column == date_column) {\n" \
 "  cmp = at-bt;\n" \
 " } else {\n" \
 "  if (sort_column == name_column) {\n" \
```

gstrauss:
I am planning on using the commits on my development branch at
https://github.com/gstrauss/lighttpd1.4/tree/mod_dirlisting
unless you have further feedback about the performance after I reverted the addition of data-value on mtime and type columns.

My remaining question is whether or not I should keep your patch addition which disables the initial sort on page load.  What are your thoughts given the Chrome performance for the initial sort?

gstrauss:
It seems that if I `tbody.innerHTML = '';\n"` *before* appending row elements to `frag`, Chrome does not have the initial 28 sec pause for the first sort, and takes 3 seconds to do the sort.

cicku:
That's a smoking gun. A quick search brought me to https://dom.spec.whatwg.org/#concept-node-pre-insert and this is what Chrome did recently: https://github.com/whatwg/dom/issues/1255. Firefox appears to have a different DOM approach and has index caching (accessability?).

- `tbody.innerHTML = ''` before appending to frag is working primarily because it goes through [ContainerNode::RemoveChildren](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/dom/container_node.cc#L1104), which then triggers `ChildrenChanged` once with `kAllChildrenRemoved`.
- Then if the rows are detached, `appendChild(row)` has no removal step.
- `tbody.appendChild(frag)` does the bulk insertion.

If enabling the initial sort, there will be DOM rewrite, also when the page loads, the user sees server order; soon JS forces a reshuffle, maybe that makes `dir-listing.sort = "disable"` look buggy?

I'd leave `c = -1` as the default and let `?C=N` URLs opt in when needed.

gstrauss:
For now, I'll keep `c = -1` even though that is a behavior change if someone is using `dir-listing.sort = "disable"` on a slow, low-memory embedded device reading from a slow USB stick and expects the (often) more powerful client running javascript to sort the directory listing once received.

Thanks for your patch!

PR #158, opened 2026-05-27
[mod_wstunnel] RFC 6455 §5 compliance

cicku:
I was investigating an unusual memory consumption (when `server.stream-request-body = 1`), noticed minor compliance issues so made some changes according to the RFC, to reject malformed input rather than silently accepting it:

- Reject 64-bit payload length with the MSB set (RFC 6455 §5.2).
- Reject Ping/Pong/Close payloads longer than 125 bytes (RFC 6455 §5.5).
- Reject fragmented control frames (RFC 6455 §5.5).
- Reject RSV1/RSV2/RSV3 bits when no extension is negotiated (RFC 6455 §5.2).

### DIFF
--- a/src/mod_wstunnel.c
+++ b/src/mod_wstunnel.c
@@ -1068,6 +1068,21 @@ static int recv_ietf_00(handler_ctx *hctx) {
 #define MOD_WEBSOCKET_FRAME_LEN16_CNT  2
 #define MOD_WEBSOCKET_FRAME_LEN63_CNT  8
 
+static int is_control_frame(const handler_ctx * const hctx) {
+    return hctx->frame.type == MOD_WEBSOCKET_FRAME_TYPE_PING
+        || hctx->frame.type == MOD_WEBSOCKET_FRAME_TYPE_PONG
+        || hctx->frame.type == MOD_WEBSOCKET_FRAME_TYPE_CLOSE;
+}
+
+static int check_control_frame_size(const handler_ctx * const hctx) {
+    if (is_control_frame(hctx)
+        && hctx->frame.ctl.siz >= MOD_WEBSOCKET_FRAME_LEN16) {
+        DEBUG_LOG_ERR("%s", "control frame size is invalid");
+        return -1;
+    }
+    return 0;
+}
+
 static int send_rfc_6455(handler_ctx *hctx, mod_wstunnel_frame_type_t type, const char *payload, size_t siz) {
     char mem[10];
     size_t len;
@@ -1159,6 +1174,14 @@ static int recv_rfc_6455(handler_ctx *hctx) {
         for (uint32_t i = 0; i < flen; ) {
             switch (hctx->frame.state) {
             case MOD_WEBSOCKET_FRAME_STATE_INIT:
+                if ((frame[i] & 0x70) != 0) {
+                    DEBUG_LOG_ERR("%s", "reserved bits are set");
+                    return -1;
+                }
+                if ((frame[i] & 0x08) && 0 == (frame[i] & 0x80)) {
+                    DEBUG_LOG_ERR("%s", "control frame is fragmented");
+                    return -1;
+                }
                 switch (frame[i] & 0x0f) {
                 case MOD_WEBSOCKET_OPCODE_CONT:
                     DEBUG_LOG_DEBUG("%s", "type = continue");
@@ -1208,12 +1231,20 @@ static int recv_rfc_6455(handler_ctx *hctx) {
                     hctx->frame.state = MOD_WEBSOCKET_FRAME_STATE_READ_MASK;
                 }
                 else if (hctx->frame.ctl.siz == MOD_WEBSOCKET_FRAME_LEN16) {
+                    if (is_control_frame(hctx)) {
+                        DEBUG_LOG_ERR("%s", "control frame size is invalid");
+                        return -1;
+                    }
                     hctx->frame.ctl.siz = 0;
                     hctx->frame.ctl.siz_cnt = MOD_WEBSOCKET_FRAME_LEN16_CNT;
                     hctx->frame.state =
                         MOD_WEBSOCKET_FRAME_STATE_READ_EX_LENGTH;
                 }
                 else if (hctx->frame.ctl.siz == MOD_WEBSOCKET_FRAME_LEN63) {
+                    if (is_control_frame(hctx)) {
+                        DEBUG_LOG_ERR("%s", "control frame size is invalid");
+                        return -1;
+                    }
                     hctx->frame.ctl.siz = 0;
                     hctx->frame.ctl.siz_cnt = MOD_WEBSOCKET_FRAME_LEN63_CNT;
                     hctx->frame.state =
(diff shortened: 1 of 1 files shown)


gstrauss:
Thanks!  Will review soon.

src/mod_wstunnel.c
```diff
@@ -1159,6 +1174,14 @@ static int recv_rfc_6455(handler_ctx *hctx) {
         for (uint32_t i = 0; i < flen; ) {
             switch (hctx->frame.state) {
             case MOD_WEBSOCKET_FRAME_STATE_INIT:
+                if ((frame[i] & 0x70) != 0) {
+                    DEBUG_LOG_ERR("%s", "reserved bits are set");
+                    return -1;
+                }
+                if ((frame[i] & 0x08) && 0 == (frame[i] & 0x80)) {
```
gstrauss: You note here that `(frame[i] & 0x08)` indicates a control frame, but `is_control_frame()` is unnecessarily more complex.

src/mod_wstunnel.c
```diff
@@ -1239,6 +1276,7 @@ static int recv_rfc_6455(handler_ctx *hctx) {
                     }
                     DEBUG_LOG_DEBUG("specified payload size=%llx",
                                     (unsigned long long)hctx->frame.ctl.siz);
+                    if (0 != check_control_frame_size(hctx)) return -1;
```
gstrauss: This check does not appear necessary since `MOD_WEBSOCKET_FRAME_STATE_READ_EX_LENGTH` state is only reached when the second byte indicates `MOD_WEBSOCKET_FRAME_LEN16` or `MOD_WEBSOCKET_FRAME_LEN63`, and the length of control frames is checked there (in your patch)

src/mod_wstunnel.c
```diff
@@ -1222,11 +1253,17 @@ static int recv_rfc_6455(handler_ctx *hctx) {
                 else {
                     DEBUG_LOG_DEBUG("specified payload size=%llx",
                                     (unsigned long long)hctx->frame.ctl.siz);
+                    if (0 != check_control_frame_size(hctx)) return -1;
```
gstrauss: This check does not appear necessary since the frame size for control frames is checked when the second byte indicates `MOD_WEBSOCKET_FRAME_LEN16` or `MOD_WEBSOCKET_FRAME_LEN63`, and the length of control frames is checked there (in your patch)

src/mod_wstunnel.c
```diff
@@ -1068,6 +1068,21 @@ static int recv_ietf_00(handler_ctx *hctx) {
...
+}
+
+static int check_control_frame_size(const handler_ctx * const hctx) {
+    if (is_control_frame(hctx)
+        && hctx->frame.ctl.siz >= MOD_WEBSOCKET_FRAME_LEN16) {
+        DEBUG_LOG_ERR("%s", "control frame size is invalid");
+        return -1;
+    }
+    return 0;
+}
```
gstrauss: See other comments, and then this helper function is no longer needed.

gstrauss (commented):
Initial comments.  I may have additional comments soon.

src/mod_wstunnel.c
```diff
@@ -1222,11 +1253,17 @@ static int recv_rfc_6455(handler_ctx *hctx) {
...
                     hctx->frame.state = MOD_WEBSOCKET_FRAME_STATE_READ_MASK;
                 }
                 i++;
                 break;
             case MOD_WEBSOCKET_FRAME_STATE_READ_EX_LENGTH:
+                if (hctx->frame.ctl.siz_cnt == MOD_WEBSOCKET_FRAME_LEN63_CNT
+                    && (frame[i] & 0x80)) {
+                    DEBUG_LOG_ERR("%s", "frame size MSB is set");
+                    return -1;
+                }
```
gstrauss: This looks incorrect to me.  Only the highest MSB of the 8-byte size is not allowed to have the high bit set.  This section of code reads and accumulates the 8 bytes one at a time and your patch applies the restriction to each byte.

cicku:
I updated with a cleaner approach, let me know if this works for you or not.

gstrauss:
Thanks.  This looks better to me.

I may have to rebase and adjust this since I have been digging into `src/mod_wstunnel.c` to reduce the code size.

> I was investigating an unusual memory consumption (when `server.stream-request-body = 1`)

Did you find anything of note?  Do you know if the unusual memory consumption in mod_wstunnel, or elsewhere?

cicku:
```
    client             lighttpd (reqbody_queue + gw.wb)              backend
→→→→→→→→→→→→→[TCP]→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→→[TCP]→→→→→→→→→→→→→
      OUT                              BUFFER                          IN
```

I'm still digging. If my backend is not fast enough to ingest the request body, or let's assume the worst case when stalled, eventually the buffer will pile up at `reqbody_queue` and OOM, since `recv_rfc_6455` does not appear to have aggregate size cap.

gstrauss:
> I'm still digging. If my backend is not fast enough to ingest the request body, or let's assume the worst case when stalled, eventually the buffer will pile up at `reqbody_queue` and OOM, since `recv_rfc_6455` does not appear to have aggregate size cap.

From a quick glance, take a look at the call to `chunkqueue_append_buffer()`.  That might need to check if the chunkqueue needs to spill into temporary files on disk.  Still, the setup call to `http_response_upgrade_read_body_unknown()` from `wstunnel_create_env()` should have resulted in some caps on read buffering.  I have to look further to see why this is or is not occurring.

Are the requests to the part of your site using mod_wstunnel all HTTP/1.1, or HTTP/2?  This might be an interaction with HTTP/2 handling.  tbd.

cicku:
I was using H/1.1 all the time because H/2 has a hardcoded 8 stream limit and it is not something I would like to use at the moment. If I understand correctly:

```
client  ->  lighttpd reqbody_queue  ->  mod_wstunnel decode  ->  gw.wb  ->  backend
 fast              RAM buffer              RFC6455 parser         RAM      slow/stalled
```

Even I can try setting `server.max-request-size` + `server.max-connections` + `write-timeout` to reduce the blast radius, within the `write-timeout` I may still be able to send fast, and the instantaneous buffer will explode (another case: backend is slowly ingesting, `write-timeout` will be refreshed and no firing, while RAM usage hiking up).

gstrauss:
@cicku please check your lighttpd.conf.  If you have configured `server.stream-request-body = 1`, then you might try changing that to `server.stream-request-body = 2` for paths handled by lighttpd mod_wstunnel.  (If you have not configured `server.stream-request-body = 1`, then lighttpd mod_wstunnel sets the equivalent of `server.stream-request-body = 2` when mod_wstunnel handles requests, but that is not done if you have explicitly configured `server.stream-request-body = 1`

gstrauss:
> I was using H/1.1 all the time because H/2 has a hardcoded 8 stream limit and it is not something I would like to use at the moment.

BTW, while the first part is true, HTTP/2 is generally faster and more efficient than HTTP/1.1, so I am curious why you think the stream limit is a limitation for you.

cicku:
Oh it is not necessarily related here - Cloudflare's pingora has configurable max concurrent streams support so part of my H/2 infra is based from there.

gstrauss:
@cicku please check your lighttpd.conf.  If you have configured `server.stream-request-body = 1`, then you might try changing that to `server.stream-request-body = 2` for paths handled by lighttpd mod_wstunnel.  (If you have not configured `server.stream-request-body = 1`, then lighttpd mod_wstunnel sets the equivalent of `server.stream-request-body = 2` when mod_wstunnel handles requests, but that is not done if you have explicitly configured `server.stream-request-body = 1`

gstrauss:
@cicku I pushed some additional commits to better address queue buffering for slow backends.  Specifically for HTTP/1.1, there was a code path used by mod_wstunnel which did not check if the in-memory queue had grown too large.  These changes are part of lighttpd git master branch.  Please test and report back.  Thank you.

PR #159, opened 2026-06-14
[mod_extforward] reject out-of-range netmask bits

cicku:
CIDR netmask bits were checked only against a lower bound, so configs such as "10.0.0.0/9999" were silently accepted.  This patch will reject bits greater than the address family maximum (/32 for IPv4, /128 for IPv6).

---

/0 is handled by `extforward.forwarder = ( "all" => "trust" )`

### DIFF
--- a/src/mod_extforward.c
+++ b/src/mod_extforward.c
@@ -317,6 +317,13 @@ static void * mod_extforward_parse_forwarder(server *srv, const array *forwarder
             free(fwd);
             return NULL;
         }
+        /* netmask bits must fit address family (IPv4 32, IPv6 128) */
+        if (nm_bits > (sock_addr_get_family(&sm->addr) == AF_INET ? 32 : 128)) {
+            log_error(srv->errh, __FILE__, __LINE__,
+              "ERROR: invalid netmask: %s", ds->key.ptr);
+            free(fwd);
+            return NULL;
+        }
         buffer_clear(&ds->value);
         /* empty is untrusted,
          * e.g. if subnet (incorrectly) appears in X-Forwarded-For */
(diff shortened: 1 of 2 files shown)


gstrauss:
> so configs such as "10.0.0.0/9999" were silently accepted.

... and the problem/impact is what?  The mod_extforward config is written by the administrator, not an untrusted entity somewhere else on the internet.  The administrator should test that the config does what the administrator wants it to do.

`sock_addr_is_addr_eq_bits()` silently reduces larger numbers of bits to the max number of bits in the address, so if you provide, say, `/40` for IPv4, all 32-bits are compared.

You're welcome to disagree with the choice, but if that is the case, please recognize the choice, state your disagreement more clearly, and why you think it is important to change.

Or is there some other problem you are trying to solve with this PR?

cicku:
No, it is just like a sanity check - to eliminate any misconfiguration if possible.

gstrauss:
> No, it is just like a sanity check - to eliminate any misconfiguration if possible.

Were that a simple sanity-check, your suggested patch should have issued warning trace.  Instead, your patch made that case an error, causing lighttpd to exit, which could break previously working configurations.

For the mod_extforward area of configuration, lighttpd is unable to validate if the distinction between, say, /23 and /24 is relevant to the administrator's network.  mod_extforward is a security boundary and administrators must test that they have configured mod_extforward to work as the administrator intends.  lighttpd can not intuit that for each environment running lighttpd.  For IPv4, /33 when /32 is meant is harmless in currently lighttpd.  On the other hand, /23 when /24 is meant could have security consequences and the administrator must detect that.

PR #162, opened 2026-06-18
[core] add lstat() support on Windows to check symlink restrictions

kali834x:
On Windows, lstat() was mapped directly to stat(), which follows symbolic links. As a result, symbolic links could not be distinguished from their targets, causing symlink checks and restrictions such as server.follow-symlink = "disable" to be ineffective.

This PR adds Windows-specific lstat() support to enable proper detection of symbolic links and directory junctions. It defines _S_IFLNK as 0xA000 and enables HAVE_LSTAT under _WIN32, maps lstat() to fs_win32_lstati64UTF8(), and implements fs_win32_lstati64UTF8() in src/fs_win32.c. The implementation detects reparse points (symbolic links and directory junctions), falls back to fs_win32_stati64UTF8() for non-reparse-point paths, and opens reparse points using FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS so that information about the link itself is retrieved rather than the target. Detected links are marked with _S_IFLNK in st_mode.

Paths ending with a trailing slash continue to fall back to normal stat() behavior, matching POSIX path resolution semantics. With these changes, Windows builds can correctly identify symbolic links and junctions, allowing existing symlink restriction checks to function as intended

### DIFF
--- a/src/fs_win32.c
+++ b/src/fs_win32.c
@@ -13,13 +13,24 @@
 /* MS filesystem API does not support UTF-8?  WTH?  write our own; not hard */
 
 #include <sys/types.h>
-#include <sys/stat.h>
-#include <direct.h>
 #include <io.h>
 
 #include <windows.h> /*(otherwise get No Target Architecture error)*/
 #include <stringapiset.h>
 #include <errno.h>
+
+#include "sys-stat.h" /*(for _S_IFLNK; include after <...> headers)*/
+
+#ifndef IO_REPARSE_TAG_SYMLINK
+#define IO_REPARSE_TAG_SYMLINK          0xA000000C
+#endif
+#ifndef IO_REPARSE_TAG_LX_SYMLINK
+#define IO_REPARSE_TAG_LX_SYMLINK       0xA000001D
+#endif
+
+#define IsReparseTagSymlink(tag) \
+    ((tag) == IO_REPARSE_TAG_SYMLINK || \
+     (tag) == IO_REPARSE_TAG_LX_SYMLINK)
 
 int fs_win32_openUTF8 (const char *path, int oflag, int pmode)
 {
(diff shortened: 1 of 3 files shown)


gstrauss:
FYI: I may not have time to look deeper into this PR for a couple weeks.

> It defines _S_IFLNK as 0xA000 and enables HAVE_LSTAT under _WIN32

Which versions of Windows support this?  (I have not yet checked.)  IIRC, Windows 10 is going out of "public" support in Oct 2026, but Windows 10 is going to be around for while longer.

> causing symlink checks and restrictions such as `server.follow-symlink = "disable"` to be ineffective.

`server.follow-symlink = "disable"` was originally an attempt to be compatible back around 2006 with a similar Apache httpd feature.  To be clear, any admin using `server.follow-symlink = "disable"` has implemented a suboptimal design of file access and web host isolation.  The admin should set up access controls in order to be able to not worry about symlinks.  `server.follow-symlink = "disable"` is inefficient and checks are not atomic.

Besides `server.follow-symlink = "disable"`, what are the reasons you are interested in having this feature?  I'll still review this PR in due time, but I'll try to find time sooner if there is a reason in addition to "because modern Windows sort of pretends to implement these POSIX-like symlink features, but not really the same way".

kali834x:
Windows support: Vista / Server 2008 and later. This does not change the minimum Windows version requirement since lighttpd already uses GetFinalPathNameByHandleW (Vista+).
Why it's needed: Windows supports symlinks and junctions, but stat() follows them, so lighttpd cannot tell whether a path is a link or the actual target. That means options like server.follow-symlink = "disable" cannot work correctly. Adding lstat() support fixes that.
Reparse points: Not all reparse points are symlinks. Things like OneDrive placeholders and NTFS deduplication also use reparse points, so checking the surrogate tag helps distinguish actual links from other reparse-point types.

src/fs_win32.c
```diff
@@ -8,19 +8,22 @@
 
 #ifdef _WIN32
 
-#include "fs_win32.h"
-
-/* MS filesystem API does not support UTF-8?  WTH?  write our own; not hard */
-
```
gstrauss, line 14: It is common practice to include the header defining the exported interfaces.

Why are you removing this include?
`fs_win32.h` includes some defines for struct sizes and it is important to include this for binary structure consistency across lighttpd files which include `fs_win32.h`.

Why are you removing the comment?  This file implements UTF-8 conversions to Windows interfaces since Windows historically supports only ASCII (`*A() funcs`) or UTF-16 (`*W() funcs`).

gstrauss:
> Why it's needed: Windows supports symlinks and junctions, but stat() follows them, so lighttpd cannot tell whether a path is a link or the actual target. That means options like server.follow-symlink = "disable" cannot work correctly. Adding lstat() support fixes that.
> Reparse points: Not all reparse points are symlinks. Things like OneDrive placeholders and NTFS deduplication also use reparse points, so checking the surrogate tag helps distinguish actual links from other reparse-point types.

You reposted this after my comment explaining the historical poor design of any system whose security relies on of using `server.follow-symlink = "disable"`.  Please take the strong hint that I do not consider anything relying on `server.follow-symlink = "disable"` a strong argument for making large code changes, especially not on Windows.

Related, do you know of real-world use of sites using lighttpd on Windows where the web document root contains symlinks, junctions, and reparse points?  Or even that the web document root is one of those rather than a subdirectory?  lighttpd currently does not recognize symlinks on Windows and uses `stat()` instead of `lstat()` works today for serving static files, as the links will be followed by the underlying Windows libraries.  Other than the bespoke `server.follow-symlink = "disable"` and the `lighty.c.readlink()` interface in lighttpd mod_magnet, could you help me to understand what your goals are with these patches and what (real-world) problems you are trying to solve?

kali834x:
Apologies for removing the header include and comment in src/fs_win32.c. That was an accidental mistake during refactoring, and I've restored both.
The goal of this PR is not only to support server.follow-symlink = "disable", but also to provide proper lstat() behavior on Windows. This helps features like mod_magnet (lighty.c.readlink()), improves symlink detection, and makes Windows behavior more consistent with other platforms.

gstrauss:
> The goal of this PR is not only to support server.follow-symlink = "disable", but also to provide proper lstat() behavior on Windows. This helps features like mod_magnet (lighty.c.readlink()), improves symlink detection, and makes Windows behavior more consistent with other platforms.

I feel like I am having a one-way conversation with an AI that has not read and understood my comments, and is regurgitating talking points which I have either already dismissed, or already argued against.

Progress is not being made.

sahvx655-wq:
A real-world issue is that Windows currently misclassifies some non-symlink reparse-point files as `DT_LNK`.
For example, App Execution Aliases such as `python.exe` and `wsl.exe` are not symlinks, but are reported as `DT_LNK` because they have the `FILE_ATTRIBUTE_REPARSE_POINT` flag set. This patch checks the actual reparse tag so only real symlinks are classified as `DT_LNK`.
It also adds proper `lstat()` behavior on Windows, allowing links to be distinguished from their targets when needed.

gstrauss:
> A real-world issue is that Windows currently misclassifies some non-symlink reparse-point files as DT_LNK.

Please ask a friend or professor you trust to help you understand what I might mean by a "real-world scenario".

I continue to wait for a description of how the behavior in lighttpd, which uses `stat()` on Windows, might have an adverse *impact* besides `server.follow-symlink = "disable"` and the `lighty.c.readlink()` interface in lighttpd mod_magnet.  (BTW, lighttpd mod_magnet `lighty.c.readlink()` need not be used; lua code can be written to access symlink functionality via a supplemental lua module, independent of lighttpd.)

For example, does lighttpd use of `stat()` on an App Execution Alias result in lighttpd not executing, say, `python.exe` when such execution is the expected/intended behavior?  That would be a real-world scenario of an impact.

kali834x:
One  impact may be in mod_webdav. On Windows, directory junctions are currently classified as DT_LNK rather than directories. This can cause WebDAV recursive operations such as PROPFIND, COPY, MOVE, and DELETE to treat a junction as a link instead of traversing it as a directory. The reparse-point classification changes address this by classifying only actual symlinks as DT_LNK.

gstrauss:
> One impact may be in mod_webdav. On Windows, directory junctions are currently classified as DT_LNK rather than directories. This can cause WebDAV recursive operations such as PROPFIND, COPY, MOVE, and DELETE to treat a junction as a link instead of traversing it as a directory. The reparse-point classification changes address this by classifying only actual symlinks as DT_LNK.

This is a weak argument.  https://wiki.lighttpd.net/mod_webdav#symlinks specifically documents "WebDAV does not have a concept of symlinks." and provides alternative guidance.

As I noted in https://github.com/lighttpd/lighttpd1.4/pull/166#issuecomment-4791584399
> BTW, in general, junctions are better avoided in the middle of the web tree; have the junction in the path before the web document root.

The way lighttpd mod_webdav currently handles this might be the desirable behavior.  How should PROPFIND, COPY, MOVE, DELETE etc treat Windows junctions?  Can a non-administrative user modify or recreate these?  I am not asking you to directly answer these questions.  I would first like to understand why lighttpd should even attempt to support this.  It sounds like a much better idea to me to not have junctions inside a WebDAV collection.

Have you tested any of this?  When I ask for "real-world" examples, I would like the examples to be *concrete examples* an include steps for reproducible observations.  Thank you.

gstrauss:
Aside: @sahvx655-wq, in your github account description: "Passinate" I think should be spelled "Passionate"

src/fs_win32.c
```diff
@@ -13,14 +13,21 @@
...
 
 #include <sys/types.h>
-#include <sys/stat.h>
+#include "sys-stat.h"
 #include <direct.h>
 #include <io.h>
 
 #include <windows.h> /*(otherwise get No Target Architecture error)*/
-#include <stringapiset.h>
 #include <errno.h>
```
gstrauss, line 20: Extreme care must be taken here.  `src/fs_win32.c` is intended to be all _WIN32 code.  To avoid possible conflicts, can "sys-stat.h" be moved to be below the last system include using "<...>", i.e. below `#include <errno.h>` ?  If using "sys-stat.h", then `<sys/stat.h>` and `<direct.h>` can be removed from `fs_win32.c`.

If `<stringapiset.h>` is not needed, then is `<windows.h>` still needed?

Note: the first `"#include "fs_win32.h"` already includes some of these headers, but it is fine to repeat them in this file where the contents of the headers are directly used.

Please carefully review if `#include "sys-stat.h"` is required or not for binary compatibility of struct sizes.  If it is required, then a `static_assert()` or `_Static_assert()` might be a good idea to make sure struct sizes match.

src/fs_win32.c
```diff
@@ -104,11 +111,11 @@ int fs_win32_readlinkUTF8 (const char *path, char *result, size_t rsz)
...
      #if 0 /*(???: should we strip "\\?\" from result?)*/
       (StrCmpNW(wbuf, L"\\\\?\\", 4) == 0)
-      ? WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS,
+      ? WideCharToMultiByte(CP_UTF8, 0,
                             wbuf+4, rd-4, result, rsz-1, NULL, NULL);
       :
      #endif
-        WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS,
+        WideCharToMultiByte(CP_UTF8, 0,
                             wbuf, rd, result, rsz-1, NULL, NULL);
```
gstrauss, line 123: Why remove `WC_ERR_INVALID_CHARS`?  Same as other PR?  Code changes without explanation?

kali834x: Reverted. That flag change was unrelated to lstat() and shouldn't have been in this PR.

src/fs_win32.c
kali834x: Moved "sys-stat.h" below the <...> includes and dropped <sys/stat.h> and <direct.h>, since both already come in via fs_win32.h at the top of the file. windows.h stays, the reparse-point calls (CreateFileW, GetFileInformationByHandleEx, GetFileInformationByHandle) need it.

On struct sizes: the layout is fixed by struct fs_win32_stati64UTF8 in fs_win32.h, which the existing (struct _stati64 *) cast in fs_win32_stati64UTF8() already depends on, so sys-stat.h is not what guarantees that here. It is only pulled in for _S_IFLNK. I can add a ck_static_assert(sizeof(struct fs_win32_stati64UTF8) == sizeof(struct _stati64)) to guard the cast if you want it.

kali834x:
On the bigger question: I don't have a concrete, tested real-world reproduction that meets the bar you're asking for, and I'll concede the follow-symlink and mod_webdav cases aren't strong enough on their own. I haven't run this against a live Windows setup with symlinks or junctions inside a document root, so I won't keep pitching scenarios I can't actually demonstrate.

The one part I'm confident is a plain correctness issue, independent of the symlink-policy discussion, is the readdir change: non-symlink reparse points (App Execution Aliases, OneDrive placeholders, junctions) are currently reported as DT_LNK because the code only tests FILE_ATTRIBUTE_REPARSE_POINT and not the reparse tag. If you'd rather keep this minimal, I'm happy to drop the lstat() addition and leave just that tag check, or close this out until I have a reproducible case to justify the lstat() work. Whatever you prefer.

gstrauss:
> The one part I'm confident is a plain correctness issue, independent of the symlink-policy discussion, is the readdir change: non-symlink reparse points (App Execution Aliases, OneDrive placeholders, junctions) are currently reported as DT_LNK because the code only tests FILE_ATTRIBUTE_REPARSE_POINT and not the reparse tag. If you'd rather keep this minimal, I'm happy to drop the lstat() addition and leave just that tag check, or close this out until I have a reproducible case to justify the lstat() work. Whatever you prefer.

Yes, there is a comment in the existing lighttpd code
`/* XXX: incomplete; need to check for IO_REPARSE_TAG_SYMLINK */`

Please submit the sys-dirent.h `readdir()` check of the tag to a new PR.
[src/sys-dirent.h](https://github.com/lighttpd/lighttpd1.4/pull/162/changes#diff-71547d816d628cb9b2ac224d7fe9cc1b4bf65b35c2a3c0616a7be85d594fafb7)

However, please do not create a new PR until you have tested it on Windows.
In this PR, you have removed `#include <stringapiset.h>` from sys-dirent.h.  Why?

gstrauss:
> On the bigger question: I don't have a concrete, tested real-world reproduction that meets the bar you're asking for, and I'll concede the follow-symlink and mod_webdav cases aren't strong enough on their own. I haven't run this against a live Windows setup with symlinks or junctions inside a document root, so I won't keep pitching scenarios I can't actually demonstrate.

I'll convert this to a draft if you would like to try to provide additional justification.

kali834x:
The stringapiset.h removal was an accident while adding the tag defines, no reason behind it. Sorry about that.

Saw you landed the readdir tag check in 05c8591 with the mount-point handling, thanks for picking that up and for the credit. Since it's in master now I haven't opened a separate PR for it. I've rebased this branch onto master, so the sys-dirent.h hunk is gone and the PR is down to a single commit with just the lstat() part. While at it I corrected the IO_REPARSE_TAG_LX_SYMLINK fallback value here to 0xA000001D to match your commit; the 0xAF000005 I had was wrong.

Draft status is fine. I don't have justification beyond what's already been discussed, so I'll leave this parked unless I can put together a concrete, tested case on Windows.

PR #164, opened 2026-06-20
[core] normalize trailing dots and spaces in Win32 URL paths

kali834x:
On Windows/Cygwin, NTFS path resolution normalizes trailing dots and spaces in path components. Intermediate components such as `admin.` are resolved as `admin`, and final components with trailing spaces or dots (e.g. `admin `) are similarly normalized. In addition, components consisting only of dots and spaces (e.g. `.. ` or `...`) may resolve to the current directory.

This difference between URL routing logic and filesystem resolution can allow attackers to craft URLs such as `/admin./index.html` that bypass URL-based access controls (e.g. `$HTTP["url"] =~ "^/admin/"`) while still resolving to valid filesystem paths on Windows.

This PR introduces Win32/Cygwin-specific URL path component normalization inside `burl_normalize()` to align URL parsing with platform path resolution behavior:

 Strips trailing dots and spaces (including `%20`) from components containing other characters.
 Normalizes dot/space-only components (e.g. `.. `, `...`, `%2e%2e%20`) to `.` (current directory equivalent).
 Preserves standard traversal components (`.` and `..`, including `%2e` and `%2e%2e`) without modification.

- Added extensive Windows-specific unit tests covering trailing dot/space cases, percent-encoded variants, and dot/space-only segments (e.g. `.. `, `...`).
- Tests executed successfully on MinGW (exit code: 0).

### DIFF
--- a/src/burl.c
+++ b/src/burl.c
@@ -288,6 +288,74 @@ static int burl_normalize_2F_to_slash (buffer *b, int qs, int flags)
 }
 
 
+#if defined(_WIN32) || defined(__CYGWIN__)
+
+/* Windows and Cygwin strip trailing '.' and ' ' from each path component when
+ * resolving a path on the filesystem, so "/admin./", "/admin%20/", and
+ * "/admin/" all map to the same directory.  src/response.c performs equivalent
+ * stripping, but only on the final component of the physical path; mirror it
+ * here for every URL path component so that URL-based access controls are not
+ * bypassed by these alternate spellings.  This runs before buffer_path_simplify()
+ * so that a component which strips down to nothing (e.g. "..." or "%2e%2e%20")
+ * becomes "." and is then collapsed along with any "." or ".." segments.
+ * ("%2e"/"%2E" have already been decoded to '.' by burl_normalize_basic_*(), and
+ *  a literal space is already encoded as "%20", so only '.' and "%20" remain.) */
+
+static int burl_win32_trailing (const char * const s, const int beg, const int end)
+{
+    if (end - beg >= 3 && s[end-3] == '%' && s[end-2] == '2' && s[end-1] == '0')
+        return 3;
+    if (end > beg && s[end-1] == '.')
+        return 1;
+    return 0;
+}
+
+
+static int burl_normalize_win32_needs (const char * const s, const int len)
+{
+    for (int i = 0; i < len; ) {
+        if (s[i] == '/') { ++i; continue; }
+        const int beg = i;
+        while (i < len && s[i] != '/') ++i;
+        const int n = burl_win32_trailing(s, beg, i);
+        if (0 == n)
+            continue;
+        if (1 == n /* trailing '.': leave "." and ".." for buffer_path_simplify() */
+            && (i - beg == 1 || (i - beg == 2 && s[beg] == '.')))
+            continue;
+        return 1;
+    }
+    return 0;
+}
+
+
+static void burl_normalize_win32_dots_spaces (buffer * const b)
+{
+    char * const s = b->ptr;
+    const int len = (int)buffer_clen(b);
+    int j = 0;
+    for (int i = 0; i < len; ) {
+        if (s[i] == '/') {
+            s[j++] = s[i++];
+            continue;
+        }
+        const int beg = j;
+        while (i < len && s[i] != '/')
+            s[j++] = s[i++];
+        if (s[j-1] == '.' /* leave "." and ".." for buffer_path_simplify() */
+            && (j - beg == 1 || (j - beg == 2 && s[beg] == '.')))
+            continue;
+        for (int n; (n = burl_win32_trailing(s, beg, j)); )
+            j -= n;
+        if (j == beg) /* component was only dots and spaces */
+            s[j++] = '.';
+    }
+    buffer_truncate(b, (size_t)j);
+}
+
+#endif /* _WIN32 || __CYGWIN__ */
+
+
 static int burl_normalize_path (buffer *b, buffer *t, int qs, int flags)
 {
     const unsigned char * const s = (unsigned char *)b->ptr;
@@ -306,12 +374,22 @@ static int burl_normalize_path (buffer *b, buffer *t, int qs, int flags)
         }
     }
 
+  #if defined(_WIN32) || defined(__CYGWIN__)
+    if (burl_normalize_win32_needs(b->ptr, qs < 0 ? used : qs))
+        path_simplify |= 2;
+  #endif
+
     if (path_simplify) {
         if (flags & HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REJECT) return -2;
         if (qs >= 0) {
             buffer_copy_string_len(t, b->ptr+qs, used - qs);
             buffer_truncate(b, qs);
         }
+
+      #if defined(_WIN32) || defined(__CYGWIN__)
+        if (path_simplify & 2)
+            burl_normalize_win32_dots_spaces(b);
+      #endif
 
         buffer_path_simplify(b);
 
--- a/src/t/test_burl.c
+++ b/src/t/test_burl.c
@@ -132,6 +132,19 @@ static void test_burl_normalize (void) {
     run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/b/."), CONST_STR_LEN("/a/b/"));
     run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/b/.."), CONST_STR_LEN("/a/"));
     run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/../b/.."), CONST_STR_LEN("/"));
+  #if defined(_WIN32) || defined(__CYGWIN__)
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/admin./index.html"), CONST_STR_LEN("/admin/index.html"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/admin...  /index.html"), CONST_STR_LEN("/admin/index.html"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/admin%20/index.html"), CONST_STR_LEN("/admin/index.html"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/admin...%20%20/index.html"), CONST_STR_LEN("/admin/index.html"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/foo.bar./baz"), CONST_STR_LEN("/foo.bar/baz"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/foo/bar.  /baz"), CONST_STR_LEN("/foo/bar/baz"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/~user./"), CONST_STR_LEN("/~user/"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/admin./index.html?q=1"), CONST_STR_LEN("/admin/index.html?q=1"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/..%20/b"), CONST_STR_LEN("/a/b"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/%2e%2e%20/b"), CONST_STR_LEN("/a/b"));
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/%2e%2e%2e/b"), CONST_STR_LEN("/a/b"));
+  #endif
     flags &= ~HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REMOVE;
 
     flags |= HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REJECT;
@@ -141,6 +154,11 @@ static void test_burl_normalize (void) {
     run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/../b"), "", (size_t)-2);
     run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/b/."), "", (size_t)-2);
     run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/b/.."), "", (size_t)-2);
+  #if defined(_WIN32) || defined(__CYGWIN__)
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/admin./index.html"), "", (size_t)-2);
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/admin...  /index.html"), "", (size_t)-2);
+    run_burl_normalize(psrc, ptmp, flags, __LINE__, CONST_STR_LEN("/a/..%20/b"), "", (size_t)-2);
+  #endif
     flags &= ~HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REJECT;
 
     flags |= HTTP_PARSEOPT_URL_NORMALIZE_QUERY_20_PLUS;


gstrauss:
Please see code which already performs some WIN32/Cygwin path normalization in `src/response.c:http_response_prepare_physical()`.  There, path normalization is performed on the physical path to the filesystem, and not on the URL since not all URLs map to filesystems.  Note the comment about the limitation that `src/response.c:http_response_prepare_physical()` might also affect path-info component of the URL.

Yes, I fully understand your point that if someone has written URL matching rules, then alternatives with trailing '`.`' and '` `' might bypass the rules.  That is, of course, among the reasons burl.c was written, and why I'll consider this patch.  

Aside: Generally speaking, when writing matching rules it is important for the admin to understand security best practices.  The admin *ought to* write rules which explicitly allow recognized paths, and then deny everything else.  Unfortunately, many people do not follow this discipline of explicit allow, default deny.

I do see that this PR adds code to run only if `HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REMOVE` flag, but that flag is enabled by default.  On such a hot code path, please consider following the code patterns in burl.c where a routine walks the URL and does nothing if no action is needed.  If action is needed, a subroutine is called with the current position, and the subroutine copies characters, again continuing to walk the URL once.  Complexity O(n).  Your patch, on the other hand, copies the query string twice, and the query string can be very long.  After your patch does `burl_normalize_win32_dots_spaces()`, instead of keeping track of the length, your patch performs an additional `strlen()` which would not be necessary if you had walked the path once, beginning to end, and kept track of the position.

I do not see where your patch extends `HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REJECT` (reject instead of remove)

Note: this was a quick look; I have not looked closely at this patch.

kali834x:
Thank you for the review and feedback. I have updated the branch to address the points raised: reworked the Win32 normalization flow to avoid extra URL/query string copies, integrated the dot/space handling into burl_normalize_path(), removed the additional strlen() by returning the new length directly, added support for HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REJECT, and added corresponding unit tests. Please let me know if you have any further suggestions.

src/burl.c
```diff
@@ -325,8 +345,144 @@ static int burl_normalize_path (buffer *b, buffer *t, int qs, int flags)
...
+            }
+        }
+    }
+    s[j] = '\0';
+    return j;
+}
+#endif
+
 __attribute_cold__
-__attribute_noinline__
```
gstrauss, line 329: why?

src/burl.c
```diff
@@ -358,6 +514,8 @@ int burl_normalize (buffer *b, buffer *t, int flags)
         qs = burl_scan_qmark(b);
     }
 
+
+
```
gstrauss: why?

src/burl.c
```diff
@@ -325,8 +345,144 @@ static int burl_normalize_path (buffer *b, buffer *t, int qs, int flags)
...
+            }
+
+            int new_len = seg_end - seg_start;
+            if (new_len == 0) {
+                if (is_dot) {
+                    s[j++] = '.';
+                } else if (is_dotdot) {
+                    s[j++] = '.';
+                    s[j++] = '.';
+                } else {
```
gstrauss: Please follow coding conventions in the file.

Adding your own coding style preferences is not appreciated.

src/burl.c
```diff
@@ -325,8 +345,144 @@ static int burl_normalize_path (buffer *b, buffer *t, int qs, int flags)
...
+                (s[seg_start] == '.' && s[seg_start+1] == '%' && s[seg_start+2] == '2' && (s[seg_start+3] == 'e' || s[seg_start+3] == 'E')) ||
+                (s[seg_start] == '%' && s[seg_start+1] == '2' && (s[seg_start+2] == 'e' || s[seg_start+2] == 'E') && s[seg_start+3] == '.')
+            )) {
+                is_dotdot = 1;
+            } else if (seg_len == 6 && 
+                s[seg_start] == '%' && s[seg_start+1] == '2' && (s[seg_start+2] == 'e' || s[seg_start+2] == 'E') &&
+                s[seg_start+3] == '%' && s[seg_start+4] == '2' && (s[seg_start+5] == 'e' || s[seg_start+5] == 'E')
+            ) {
+                is_dotdot = 1;
+            }
```
gstrauss: Doesn't burl.c take care of normalizing "%2e" and "%2E" to "." elsewhere, if an appropriate flag is set?  Why are you repeating here?  Should your code run after that normalization code has been run?

src/burl.c
```diff
@@ -306,13 +312,27 @@ static int burl_normalize_path (buffer *b, buffer *t, int qs, int flags)
         }
     }
 
+  #if defined(_WIN32) || defined(__CYGWIN__)
+    int win32_sanitize = burl_win32_needs_normalize(b->ptr, len);
+    if (win32_sanitize) {
+        path_simplify = 1;
+    }
+  #endif
+
```
gstrauss, line 381: Saving len here is not the pattern/optimization I was seeking.  If the code   If the path does not otherwise need to be modified (`path_simplify == 0`), then you might want to check if it needs to be modified (`path_simplify != 0`) on `_WIN32` or `__CYGWIN__`.  You might set a bit flag `path_simplify |= 2` instead of using a separate variable `win32_sanitize`.  Hint: if you find yourself modifying code not within a `#if defined(_WIN32) || defined(__CYGWIN__)`, there should be justification explicitly explained to me.

When `path_simplify == 0` after the above, that is the fast path: no changes needed, and hopefully that is (by far) the common code path taken.

When `path_simplify != 0`, this is where evasive action is taken, and there is a larger cost to modify the path.  Routines operating on the path to modify it should try to minimize copies and scan, but work still needs to be done.  One technique used here is to copy the path-to-be-modified into a temporary buffer and apply modifications into result buffer, which is the same as the source buffer.  Under the covers, any buffer allocation resizing occurs on the result buffer.

gstrauss (commented):
Quick review; not complete.   I was slowed down by seemingly random changes (removing a function attribute without explanation) and code inconsistencies (sloppily adding two blank lines for no apparent reason), and code style clashes (in this case, cuddled 'else's do not match any other code in the file).

While the content of the code is more important, the code smells -- whether accidental or not -- indicate that more time needs to be sunk into code reviews, questioning everything and lowering the assessment of code quality.

src/burl.c
kali834x: Accidental drop, no reason behind it. Restored __attribute_noinline__.

src/burl.c
kali834x: Also accidental. Removed.

src/burl.c
kali834x: Fixed. Rewrote the new functions to follow the existing style in the file (no cuddled else, etc.).

src/burl.c
kali834x: Correct, and that was the right instinct. burl_normalize_basic_*() already decodes %2e/%2E to '.' before burl_normalize_path() runs (since '.' is unreserved), so that branch was redundant. Removed it. The win32 code now only deals with literal '.' and trailing space, which stays as %20 (space is required-encoded, so it isn't decoded earlier).

src/burl.c
kali834x: Reworked along these lines. The unrelated non-win32 change (hoisting len out of the loop) is reverted. The win32 check now sets path_simplify |= 2 instead of a separate variable, so path_simplify == 0 stays the fast path with no extra work. When action is needed, the path is rewritten in a single in-place pass before buffer_path_simplify(), and the routine returns the new length so there's no follow-up strlen(). DOTSEG_REJECT rejects these paths too.

kali834x:
Pushed a reworked version, squashed to a single commit.

Main changes from the review: dropped the redundant %2e/%2E handling (already decoded by burl_normalize_basic_*() before this runs), reverted the unrelated edits (function attribute, stray blank lines), matched the file's coding style, and moved the win32 detection onto a path_simplify bit so the no-op case stays on the existing fast path and the rewrite is a single in-place pass. Builds clean and the burl unit tests pass, including the win32 cases.

gstrauss:
If `burl_normalize_win32_dots_spaces()` is moved after `buffer_path_simplify()`, would you need to check for `"."` and `".."` in `burl_normalize_win32_dots_spaces()`?  Add a comment that `buffer_path_simplify()` has already removed any  `"."` or `".."` path segments.

Can `burl_normalize_win32_dots_spaces()` simply take `(buffer *)` and call `buffer_truncate()` before returning?

Inside `burl_normalize_win32_dots_spaces()`, instead of `memmove()` of the string segment, would it make sense on this "modification" code path to copy char-by-char and back up a few chars when needed?  Also, if a segment is ever all dots, you'll want to remove the segment, including the '/' to avoid double-slashes in the normalized result.  This is necessary if `burl_normalize_win32_dots_spaces()` is moved after `buffer_path_simplify()`.

src/burl.c
```diff
@@ -288,6 +288,76 @@ static int burl_normalize_2F_to_slash (buffer *b, int qs, int flags)
...
+ * here for every URL path component so that URL-based access controls are not
+ * bypassed by these alternate spellings.
+ * ('.' and '..' are preserved for buffer_path_simplify(); "%2e"/"%2E" have
+ *  already been decoded to '.' by burl_normalize_basic_*(); ' ' remains encoded
+ *  as "%20" since space must be percent-encoded.) */
+
+static int burl_win32_trailing (const char * const s, const int beg, const int end)
+{
+    if (end > beg && (s[end-1] == '.' || s[end-1] == ' '))
+        return 1;
```
gstrauss: If `s[end-1] == '.'`, then check if "." or ".." (and skip) instead of always checking for "." or ".." in `burl_normalize_win32_needs()`.

src/burl.c
```diff
@@ -288,6 +288,76 @@ static int burl_normalize_2F_to_slash (buffer *b, int qs, int flags)
...
+ * stripping, but only on the final component of the physical path; mirror it
+ * here for every URL path component so that URL-based access controls are not
+ * bypassed by these alternate spellings.
+ * ('.' and '..' are preserved for buffer_path_simplify(); "%2e"/"%2E" have
+ *  already been decoded to '.' by burl_normalize_basic_*(); ' ' remains encoded
+ *  as "%20" since space must be percent-encoded.) */
+
+static int burl_win32_trailing (const char * const s, const int beg, const int end)
+{
+    if (end > beg && (s[end-1] == '.' || s[end-1] == ' '))
```
gstrauss: `|| s[end-1] == ' '` will have been encoded to `%20` by `burl_normalize_basic_*()`, so the check is unnecessary.

gstrauss:
Your patch adds code to remove/reject trailing `.` and `%20` on _WIN32 and Cygwin under `HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REMOVE`/`HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REJECT`.

Should that be a separate option?  It is not a dot or dot-dot path segment.

I have not made up my mind yet, but I am leaning towards a separate option (for _WIN32 and Cygwin, and a no-op on other platforms) because URLs are not solely mapped to the filesystem, and these characters may legitimately be present at the ends of path segments in URLs which are not intended to be mapped to the filesystem.

src/burl.c
kali834x: Done. burl_normalize_win32_needs() now reads the trailing run first and only does the "."/".." check when that run is a single trailing '.', since that's the only shape a real dot-segment can take. Same ordering in the rewrite pass.

src/burl.c
kali834x: Right, removed. A literal trailing space is already encoded to "%20" by burl_normalize_basic_*() before this runs, so burl_win32_trailing() now only checks for '.' and "%20".

kali834x:
Pushed an update.

burl_normalize_win32_dots_spaces() now takes the buffer and does its own buffer_truncate(), and the rewrite copies char by char and backs up over the trailing run instead of memmove'ing the segment.

On moving it after buffer_path_simplify(): I looked at that. The strip is greedy over both '.' and "%20", so a component can never reduce down to a bare "." or ".." that would still need collapsing. It either keeps a non-dot/space tail or empties out entirely, so the order doesn't change the result. I left it before simplify because an emptied component can then just become "." and the existing simplify pass folds it away, which avoids the "remove the component and a '/'" handling (and the "/.../" -> "/" style edge cases) that running after would need. Happy to flip it if you'd rather drop the dot handling instead.

On a separate option: agreed, that's cleaner. Trailing '.'/space only matter once the path maps to the filesystem and can be legitimate on URLs that never do, so folding it into dotseg-remove/reject overreaches. I can put it behind its own _WIN32/Cygwin parseopt that's a no-op elsewhere. The open call is the default: on by default keeps the bypass closed out of the box, opt-in avoids touching non-filesystem URLs. Which would you prefer and I'll wire it up.

gstrauss:
> On moving it after buffer_path_simplify(): I looked at that. The strip is greedy over both '.' and "%20", so a component can never reduce down to a bare "." or ".." that would still need collapsing. It either keeps a non-dot/space tail or empties out entirely, so the order doesn't change the result.

'..' needs to be reserved so that `buffer_path_simplify()` running afterwards would remove the '..' and the prior path segment.

> I left it before simplify because an emptied component can then just become "." and the existing simplify pass folds it away, which avoids the "remove the component and a '/'" handling (and the "/.../" -> "/" style edge cases) that running after would need.

That reads as somewhat "action at a distance" and suggests that a comment should have explained that dependency.  

In any case, I think you should reconsider.  buffer_path_simplify() removes '.' and '..' segments.  If the _WIN32 additional normalization steps for trailing '.' and ' ' occurred after, then the _WIN32 additional normalization steps would then not have to check for and special-case '..', and the code would be better isolated and slightly simpler.

gstrauss:
I won't have time to return to these PRs for at least a week.

However, I'll checkpoint with a note that this PR is incomplete.

In `burl_normalize()`, `HTTP_PARSEOPT_URL_NORMALIZE_PATH_BACKSLASH_TRANS` does not currently translate `%5C` to `%2F` and perhaps it should.  That flag is not enabled by default.  URLs are valid with these encodings and might not be mapped to the filesystem.  If mapped to the filesystem, then full normalization is recommended.  If not mapped to the filesystem, then normalizing the URLs for the filesystem would CORRUPT the URL.

`buffer_path_simplify()` might be an additional location to check for trailing '.' and ' ' on _WIN32, as `buffer_path_simplify()` is also called on paths after `buffer_urldecode_path()`.  Possibly better would be to review locations where `buffer_path_simplify()` is called, and if those are filesystem paths (instead of URLs), add additional normalization for _WIN32.  However, sometimes a url path is later appended to physical path, and would need the additional _WIN32 filesystem normalization applied then.

Just some checkpoint notes of incomplete thoughts.

kali834x:
You're right, I had that backwards. Stripping can reveal a ".." that still needs collapsing: "/a/..%20/b" is "/a/.. /b" on Windows, the trailing space drops, and it resolves to "/b", not "/a/b". So the current code (and that test) is wrong, it folds "..%20" down to "." instead of reserving the "..".

That's actually why I think the strip needs to stay before buffer_path_simplify() rather than after. simplify only acts on literal "." and "..", so it has to see the revealed ".." to drop it and the prior segment. If the strip runs after simplify, "..%20" is still "..%20" when simplify passes, and the ".." we reveal afterwards gets left in the path unless we simplify a second time. Running before lets the single existing simplify pass handle it. Downside is the ordering dependency you flagged, which I'll spell out in a comment so it isn't action-at-a-distance.

If you'd rather have it after simplify, the correct way is a second simplify pass after the strip. Happy to go that way instead, just say which you prefer and I'll fix the reveal logic + tests to match.

On the separate option: agreed, I'll gate it behind its own _WIN32/Cygwin parseopt that's a no-op elsewhere. The only open call is the default (on by default keeps the bypass closed out of the box vs opt-in to avoid touching non-filesystem URLs). I'll wire it up once you pick.

For %5C -> %2F under BACKSLASH_TRANS and the other buffer_path_simplify() callers, those read as wider than this PR, and %5C can corrupt URLs that never map to the filesystem like you said. I'd keep this one to the URL-path strip and do those separately unless you'd rather fold them in. No rush, I know you're tied up for a bit.

gstrauss:
> Stripping can reveal a ".." that still needs collapsing: "/a/..%20/b" is "/a/.. /b" on Windows, the trailing space drops, and it resolves to "/b", not "/a/b".

Good point.  It might make sense to do this before `buffer_path_simplify()`, or to perform the equivalent of `buffer_path_simplify()` while normalizing Windows inanities.

I think these Windows inanities should be a separate flag, but should be done along with `HTTP_PARSEOPT_URL_NORMALIZE_PATH_DOTSEG_REMOVE`.

gstrauss:
You keep mentioning the vague phrase "the bypass".  While I understand to what you are referring, the average reader will not.  Also, I think you have misunderstand the potential scope.

In the lighttpd configuration, mod_rewrite (`url.rewrite` et al) and mod_redirect (`url.redirect`) operate on the percent-encoded URL.  On the other hand, the lighttpd.conf conditional blocks such as `$HTTP["url"] =~ "..." { ... }` operate on internal `r->url.path` which has been percent-decoded.  `r->url.path` may be used by many modules as a virtual path (and might not be present in the filesytem), and also may be used to construct a filesystem path if the module handling the request uses the filesystem.

In the early phases of request processing, lighttpd can no a priori know whether or not the URL is intended as a virtual path or is to be mapped into a filesystem path.  *YOU* might assume everything goes to the filesystem and therefore be concerned about a malicious URL bypassing *YOUR* mod_rewrite or mod_redirect rules.  This is not a bypass in lighttpd.  lighttpd is a web server and handles the virtual URL path.  lighttpd mod_staticfile and a few other lighttpd modules might ultimately then map the URL path to the filesystem.  Other lighttpd modules might not.

Windows behavior remains obnoxious and necessitates multiple passes to normalize the URL and filesystem paths at multiple places during request processing.  `%2F` and `%5C` might be intentionally encoded in the virtual URL path.  However, if mapped to the filesystem, these now both become relevant and equivalent in _WIN32 paths.

gstrauss:
Aside: there may be places in lighttpd which have to take precautions to preserve leading `//` or `\\` in Windows paths since normalization might reduce them to a single `/`.

gstrauss:
Windows filesystems (at least some) support trailing '.' or ' '.  It appears that Windows command line tools are the ones which remove trailing '.' and ' '.

=> Have you tested and verified on Windows that lighttpd is returning the same file if a URL contains extra trailing '.' or ' '.  Is lighttpd mapping it to the same file?

kali834x:
On the same-file question: the Win32 fs layer in src/fs_win32.c opens and stats through _wopen/_wstati64/CreateFileW on plain paths with no "\\?\" prefix, so those calls go through Win32 path canonicalization, which strips trailing "." and " " from the final component. That is why "/admin./index.html" and "/admin/index.html" land on the same file, and intermediate components get the same treatment as the path is walked. What I have actually run is the burl unit tests, which exercise the normalization logic; I have not done a live file round trip on a Windows box, so I would frame the mapping as expected from the API behavior and confirmed in the unit logic rather than observed end to end. I can set up a MinGW run that creates "admin\" and requests "/admin./" to confirm the round trip if you want it nailed down.

Fair point on "the bypass", that wording was sloppy. To restate it in the right terms: lighttpd serves the virtual, percent-decoded r->url.path, and that is legitimate since not everything maps to the filesystem. The narrow case is only when an admin writes a $HTTP["url"] match and a later module (mod_staticfile and friends) maps that same path onto a Windows filesystem, where the trailing "." / " " spellings collapse. %2F and %5C can absolutely be intentional in a virtual path and only become equivalent once mapped to a Win32 path, so this normalization only makes sense as filesystem-facing, not as a blanket URL rewrite.

That is the case for the separate flag, agreed. I will move it off DOTSEG_REMOVE/REJECT onto its own _WIN32/Cygwin parseopt that is a no-op elsewhere and runs alongside DOTSEG_REMOVE, kept before buffer_path_simplify() so a revealed ".." (e.g. "/a/..%20/b") is still collapsed by simplify together with the prior segment, with a comment spelling out that ordering dependency.

%5C->%2F under BACKSLASH_TRANS, the other buffer_path_simplify() callers, and preserving leading "//" / "\\" all read as wider than this patch and can corrupt virtual URLs that never touch the filesystem, so I would keep those as separate follow-ups unless you would rather fold them in. No rush on my end.

gstrauss:
> I have not done a live file round trip on a Windows box, so I would frame the mapping as expected from the API behavior and confirmed in the unit logic rather than observed end to end. I can set up a MinGW run that creates "admin" and requests "/admin./" to confirm the round trip if you want it nailed down.

This is a prerequisite for any PR.  I will not spend an iota of time on this until you have verified the *actual* behavior, using lighttpd, and documented examples of paths tried and how the results *actually* map to the same file, and precisely when they do, and when they do not.

kali834x:
Fair, that's the right bar for a security change and I shouldn't have leaned on the fs API behavior alone. I don't have observed-on-Windows results yet, so I'll hold off asking for more of your time until I do.

Plan: build lighttpd under MinGW and serve a tree with both a normal `admin\index.html` and files created via the `\\?\` prefix that actually keep a trailing '.' / ' ' in the name, then request the plain, trailing-dot, trailing-space, `%20`, and mixed spellings and record for each which ones open the same file and which 404. I'll post that table, including precisely where the mapping breaks, before pinging you again.

gstrauss:
I'll note that in `src/response.c:http_response_prepare_physical()`, where on _WIN32 and Cygwin the final path segment has trailing `.` and ` ` removed, this occurs before the URL is checked against the filesystem for path-info, and the modified path is stored in `r->physical.rel_path` and used to construct `r->physical.path`.  You may want to temporarily disable this code during your testing.  You might also want to annotate the code to see what lighttpd gets as it calls `stat` via `stat_cache_get_entry()` to determine path-info in `src/response.c:http_response_physical_path_check()`.

kali834x:
Useful pointer. For the raw filesystem runs I'll disable the final-segment strip in http_response_prepare_physical() so it doesn't mask what the fs itself does, and add logging around the stat_cache_get_entry() calls in http_response_physical_path_check() to capture the exact paths lighttpd passes to stat, including the path-info walk. It also means the matrix should separate final-component spellings (which that code already normalizes today) from intermediate ones like /admin./index.html that currently reach the fs as-is. I'll cover both in the table.

PR #165, opened 2026-06-23
mod_alias: fix alias.url configuration order check bypass

kali834x:
This PR fixes a bug in mod_alias_check_order() where alias shadowing warnings could be skipped when unrelated aliases appeared between a prefix alias and a shadowed alias.

For example:

alias.url += (
    "/blog" => "/var/www/blog",
    "/test" => "/var/www/test",
    "/blog/images" => "/var/www/images"
)

In this case, /blog/images is shadowed by /blog, but the warning was not emitted.

The existing implementation relied on assumptions about contiguous ordering in the sorted alias array. However, alias entries are stored in insertion order, and the sorted view does not guarantee that prefix-related entries remain contiguous. As a result, the validation logic could terminate early and miss shadowed aliases.

This change replaces the ordering-based optimization with a direct prefix comparison against previously defined aliases in insertion order, ensuring that all shadowed aliases are detected regardless of intervening entries.

A new test, test_mod_alias_check_order(), has been added to verify that valid alias configurations pass validation and that shadowed aliases correctly trigger configuration errors.

### DIFF
--- a/src/fs_win32.c
+++ b/src/fs_win32.c
@@ -8,18 +8,21 @@
 
 #ifdef _WIN32
 
-#include "fs_win32.h"
-
-/* MS filesystem API does not support UTF-8?  WTH?  write our own; not hard */
-
 #include <sys/types.h>
-#include <sys/stat.h>
+#include "sys-stat.h"
 #include <direct.h>
 #include <io.h>
 
 #include <windows.h> /*(otherwise get No Target Architecture error)*/
-#include <stringapiset.h>
 #include <errno.h>
+
+#ifndef IO_REPARSE_TAG_LX_SYMLINK
+#define IO_REPARSE_TAG_LX_SYMLINK 0xAF000005
+#endif
+
+#define IsReparseTagSymlink(tag) \
+    ((tag) == IO_REPARSE_TAG_SYMLINK || \
+     (tag) == IO_REPARSE_TAG_LX_SYMLINK)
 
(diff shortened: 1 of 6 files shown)


gstrauss:
There seems to be excess, unrelated commits as part of this PR.

The commit "mod_alias: fix alias.url configuration order check bypass" seems to be a nice simpification, thank you.

gstrauss:
Tests should not output expected log trace.
```
--- a/src/t/test_mod_alias.c
+++ b/src/t/test_mod_alias.c
@@ -110,11 +110,8 @@ static void test_mod_alias_check(void) {
 static void test_mod_alias_check_order(void) {
     server srv;
     memset(&srv, 0, sizeof(server));
-    log_error_st errh;
-    memset(&errh, 0, sizeof(log_error_st));
-    errh.mode = FDLOG_FD;
-    errh.fd = 2; /* STDERR_FILENO */
-    srv.errh = &errh;
+    srv.errh = fdlog_init(NULL, -1, FDLOG_FD);
+    srv.errh->fd = -1; /* (disable) */
 
     array * const aliases = array_init(3);
 
@@ -133,6 +130,7 @@ static void test_mod_alias_check_order(void) {
     assert(0 == mod_alias_check_order(&srv, aliases));
 
     array_free(aliases);
+    fdlog_free(srv.errh);
 }
 
 void test_mod_alias (void);
```

PR #166, opened 2026-06-23
fix(core): readlink on Windows should read symlink/junction target without resolving it

kali834x:
On Windows, fs_win32_readlinkUTF8() resolved symlink and junction targets using GetFinalPathNameByHandleW(), which caused relative links to be converted to absolute paths and prevented reading broken links. This change opens reparse points directly and retrieves the stored target using FSCTL_GET_REPARSE_POINT, allowing readlink() to correctly return symlink and junction targets without resolving them, including relative and broken links, matching POSIX semantics more closely.

### DIFF
--- a/src/fs_win32.c
+++ b/src/fs_win32.c
@@ -18,6 +18,7 @@
 #include <io.h>
 
 #include <windows.h> /*(otherwise get No Target Architecture error)*/
+#include <winioctl.h>
 #include <stringapiset.h>
 #include <errno.h>
 
@@ -77,6 +78,30 @@ int fs_win32_stati64UTF8 (const char *path, struct fs_win32_stati64UTF8 *st)
     return 0;
 }
 
+// Custom reparse data buffer structure to avoid dependency on WDK/winnt.h differences
+typedef struct {
+    ULONG  ReparseTag;
+    USHORT ReparseDataLength;
+    USHORT Reserved;
+    union {
+        struct {
+            USHORT SubstituteNameOffset;
+            USHORT SubstituteNameLength;
+            USHORT PrintNameOffset;
+            USHORT PrintNameLength;
+            ULONG  Flags;
+            WCHAR  PathBuffer[1];
+        } SymbolicLinkReparseBuffer;
+        struct {
+            USHORT SubstituteNameOffset;
+            USHORT SubstituteNameLength;
+            USHORT PrintNameOffset;
+            USHORT PrintNameLength;
+            WCHAR  PathBuffer[1];
+        } MountPointReparseBuffer;
+    } u;
+} fs_win32_reparse_data_buffer_t;
+
 int fs_win32_readlinkUTF8 (const char *path, char *result, size_t rsz)
 {
     WCHAR wbuf[4096];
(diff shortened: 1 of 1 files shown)


gstrauss:
This PR adds some expensive library and system calls in an effort for pedantic, esoteric features of Windows.  In your PR, `CreateFileW()` now does a full open with `FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE` instead of `0`, and a call to `DeviceIoControl()` is added.  Not your fault, but this is a gross amount of additional work and many lines of code to attempt to imitate POSIX `readlink()` semantics.

In this PR, why does the call to `WideCharToMultiByte()` remove the `WC_ERR_INVALID_CHARS` flag?

BTW, in general, junctions are better avoided in the middle of the web tree; have the junction in the path before the web document root.

FYI: before I move further on this PR, I need to spend some time to better understand the complexity of Windows reparse points.

kali834x:
Good points.

On WC_ERR_INVALID_CHARS: there was no reason to drop it, that was an accidental change on my part. I've put it back.

On the share mode: also not needed here, so I reverted CreateFileW back to the original (share mode 0, no security attrs). The only added flag now is FILE_FLAG_OPEN_REPARSE_POINT.

The DeviceIoControl(FSCTL_GET_REPARSE_POINT) call is harder to avoid. GetFinalPathNameByHandleW always resolves the link, so it can't return a relative target and fails outright on a broken link. FSCTL_GET_REPARSE_POINT is the documented way to read the stored target as-is, which is what readlink() should return. I agree it's a lot of code for a niche corner of Windows behavior.

Agreed on keeping junctions before the document root rather than inside the tree.

No rush on the reparse-point review, take the time you need.

src/fs_win32.c
```diff
@@ -86,36 +111,99 @@ int fs_win32_readlinkUTF8 (const char *path, char *result, size_t rsz)
...
+    if (h == INVALID_HANDLE_VALUE) {
+        DWORD err = GetLastError();
+        errno = (err == ERROR_FILE_NOT_FOUND || err == ERROR_PATH_NOT_FOUND) ? ENOENT : EINVAL;
+        return -1;
+    }
+
+    union {
+        fs_win32_reparse_data_buffer_t rdb;
+        BYTE dummy[16384]; /* MAXIMUM_REPARSE_DATA_BUFFER_SIZE */
+    } u;
```
gstrauss, line 126: To keep stack size in check, can you re-use wbuf as the previous code did?  lighttpd uses a 4k buffer for filenames.  Yes, Windows supports longer filenames, but that does not mean it is a good idea or necessary.

Re-using wbuf would mean changing wbuf to a union for alignment purposes with `fs_win32_reparse_data_buffer_t`.  Please keep the size at 4k unless you have data that you can share on how using paths larger than 4k is somehow frequent and important to support.  That means also checking for `ERROR_MORE_DATA` and instead of handling it, set `errno = ENOSPC`.

src/fs_win32.c
```diff
@@ -86,36 +111,99 @@ int fs_win32_readlinkUTF8 (const char *path, char *result, size_t rsz)
...
+            target_path = &u.rdb.u.MountPointReparseBuffer.PathBuffer[printOffset];
+            target_len = printLen;
+        } else if (subLen > 0) {
+            target_path = &u.rdb.u.MountPointReparseBuffer.PathBuffer[subOffset];
+            target_len = subLen;
+            if (target_len >= 4 && target_path[0] == L'\\' && target_path[1] == L'?' && target_path[2] == L'?' && target_path[3] == L'\\') {
+                target_path += 4;
+                target_len -= 4;
+            }
+        }
```
gstrauss, line 179: Previous code contained the comment: `/*(???: should we strip "\\?\" from result?)*/`
You chose to implement that.  Why?  Because you could?  So could I, but I chose not to do so and instead to leave a note asking the question.

Please answer the question before implementing more work.  Have you tested lighttpd without doing the translation?  Given the places where lighttpd currently might use `fs_win32_readlinkUTF8`, I think it more correct to leave the `\??\`, as the prior code did.  Removing it would be okay for `\??\C:\...`, but is probably not be okay for other paths, such as `\??\UNC\...`

src/fs_win32.c
```diff
@@ -86,36 +111,99 @@ int fs_win32_readlinkUTF8 (const char *path, char *result, size_t rsz)
...
         return -1;
     }
-    /*(???: should we translate '\\' to '/' in result?)*/
     result[mlen] = '\0';
+
+    for (int i = 0; i < mlen; i++) {
+        if (result[i] == '\\') {
+            result[i] = '/';
+        }
+    }
```
gstrauss, line 206: Similar to above comment, is it *necessary* to translate `\` to `/` in what lighttpd returns from `fs_win32_readlinkUTF8`?  While not necessary, it might be a convenience for consistency when used by lighttpd mod_magnet `lighty.c.readlink()`, so I'll consider it.

gstrauss (commented):
Please be more careful with your PRs.  There have been *multiple* times in recent PRs where you have made "accidental" changes, removals, etc.

Every code change should be intentional and justifiable.

gstrauss:
References:
https://learn.microsoft.com/en-us/windows/win32/fileio/reparse-points
https://learn.microsoft.com/en-us/windows/win32/fileio/reparse-point-tags
https://learn.microsoft.com/en-us/windows/win32/fileio/reparse-points-and-file-operations

gstrauss:
Why are you checking for `IO_REPARSE_TAG_MOUNT_POINT`?  If you are checking for `IO_REPARSE_TAG_MOUNT_POINT`, then why are you not checking for `IO_REPARSE_TAG_NFS` or other types?

Why are you checking for anything beyond `IO_REPARSE_TAG_SYMLINK`?

Conversely, why are you checking the type at all instead of using Windows to resolve the path?

=> What are you *intentionally* trying to do?  What is your intention?

gstrauss:
Can this PR be reduced to adding `FILE_FLAG_OPEN_REPARSE_POINT` to the call to `CreateFileW`?  What does all the rest of the code do beyond that?  Is `GetFinalPathNameByHandleW` insufficient?  (I have not tested or looked further.)

gstrauss:
I wrote:
>> Can this PR be reduced to adding `FILE_FLAG_OPEN_REPARSE_POINT` to the call to `CreateFileW`? What does all the rest of the code do beyond that? Is `GetFinalPathNameByHandleW` insufficient? (I have not tested or looked further.)

You previously wrote:
> GetFinalPathNameByHandleW always resolves the link, so it can't return a relative target and fails outright on a broken link.

I see that you already answered my question.  Thanks.

https://web.archive.org/web/20190127160806/https://blogs.msdn.microsoft.com/oldnewthing/20100212-00/?p=14963

Edit: I still think you should test `CreateFileW` with `FILE_FLAG_OPEN_REPARSE_POINT`, and using `GetFinalPathNameByHandleW` to confirm.

src/fs_win32.c
gstrauss: Note: `\\?\` is different from `\??\`

gstrauss:
Here are some references to other code with open source licenses, though different from lighttpd BSD 3-Clause.

https://github.com/scanset/RatchetBox/blob/main/Windows/cpp/kb/win32/FileIO/reparse-point-operations.md
https://github.com/googleprojectzero/symboliclink-testing-tools/blob/00c0fe4c/CommonUtils/ReparsePoint.cpp
https://github.com/libuv/libuv/blob/v1.x/src/win/fs.c#L186
https://github.com/JFLarvoire/SysToolsLib/blob/master/C/MsvcLibX/src/readlink.c

---

At this point, I am not convinced that it is worthwhile to add this level of Windows symlink support to lighttpd.  If this functionality is needed by someone using lua with lighttpd mod_magnet, then it is still possible for them to add a lua module which contorts to Windows semantics to approximate `readlink()`.

gstrauss:
Another reference: Cygwin https://github.com/cygwin/cygwin/blob/main/winsup/cygwin/path.cc `check_reparse_point_target()` uses a different set of routines.

It is true that the current lighttpd implementation of `fs_win32_readlinkUTF8()` operates more closely to `realpath()` than `readlink()`.  

However, this `readlink()` implementation is currently used only in lighttpd mod_magnet to implement lighttpd lua interface `lighty.c.readlink()`.

As I mentioned above, I do not see the value in spending excessive amount of time trying to improve this for `_WIN32`.  A lua C module could provide what is needed if special support for interfacing with Windows is needed for symlinks, junctions, mount points, etc.

src/fs_win32.c
kali834x: Makes sense. If this stays I'll fold the reparse buffer into wbuf as a union so it's a single 4k stack buffer rather than adding the separate 16k one, and map ERROR_MORE_DATA to ENOSPC instead of trying to grow. I don't have data showing targets past 4k matter here, so 4k is the right cap.

src/fs_win32.c
kali834x: I stripped it because the substitute name comes back as an NT path like `\??\C:\target` and I wanted a plain path out. But you're right that it only holds for the drive-letter case; lopping four chars off `\??\UNC\server\share` leaves a broken path. If this stays I'll leave the `\??\` in place like the prior code did. And noted, `\\?\` and `\??\` aren't the same prefix.

src/fs_win32.c
kali834x: Not necessary. It was for consistency with the forward-slash paths lighttpd hands back elsewhere, but the raw target keeps backslashes on Windows and readlink() shouldn't rewrite them. I'll drop the translation unless you'd rather keep it for the mod_magnet caller.

kali834x:
On intention: the goal was to make fs_win32_readlinkUTF8() behave like POSIX readlink() rather than realpath(), so it returns the stored target verbatim, including relative targets and broken/dangling links. GetFinalPathNameByHandleW always resolves to a final absolute path and fails outright on a broken link, so it can't give that. Reading the raw reparse buffer is what forces the tag branch: the symlink and mount-point buffers have different layouts, so the type check is a consequence of parsing the buffer, not a feature in itself.

On the tag check being inconsistent: you're right. I singled out IO_REPARSE_TAG_MOUNT_POINT because junctions are the usual way directory links appear on Windows, but readlink() on POSIX only applies to symlinks, and handling MOUNT_POINT while ignoring NFS and the rest is arbitrary. The clean version is IO_REPARSE_TAG_SYMLINK only, returning EINVAL for anything else, which is what readlink() does.

On the bigger question: I take the point. This is a lot of code for a path only reachable through mod_magnet's lighty.c.readlink(), and the existing realpath-like behavior is fine for most uses. I'm happy to pare it down to the minimal SYMLINK-only form, or to close it and leave the niche cases to a lua C module. Your call. I'd want to test the FILE_FLAG_OPEN_REPARSE_POINT plus GetFinalPathNameByHandleW combination you suggested before reshaping it, and I'm not set up to test that on Windows at the moment, so I'll hold off on more code until you've decided whether it's worth carrying. Point taken on keeping every change deliberate.

gstrauss:
I could remove readlink support on Windows from lighttpd mod_magnet with this:
```
--- a/src/mod_magnet.c
+++ b/src/mod_magnet.c
@@ -1619,6 +1619,7 @@ static int magnet_header_tokens(lua_State *L) {
 }
 
 static int magnet_readlink(lua_State *L) {
+  #ifndef _WIN32
     const char * const path = luaL_checkstring(L, 1);
     buffer * const b = magnet_tmpbuf_acquire(L);
     ssize_t rd = readlink(path, b->ptr, buffer_string_space(b));
@@ -1627,6 +1628,10 @@ static int magnet_readlink(lua_State *L) {
     else
         lua_pushnil(L);
     magnet_tmpbuf_release(b);
+  #else
+    errno = ENOTSUP;
+    lua_pushnil(L);
+  #endif
     return 1;
 }
 
```

gstrauss:
Given the options of this PR, the current lighttpd code which on Windows is more akin to `realpath()` (with some utility), and removing support for readlink on Windows in lighttpd mod_magnet, at this juncture, I am going to retain the current lighttpd behavior.  Without demonstrating real-world need, and with demonstrating an incomplete understanding of the (bordering absurd) complexities of Windows, this PR is not ready.  Someone needing to interface with these features of the Windows filesystem should do so exactly as they need to do so with a lua module.

kali834x:
That's fair. Agreed it's a lot of code for a path only reachable through mod_magnet's readlink, and I'd rather not land half-understood reparse-point handling in core. Retaining the current behavior makes sense, and anyone needing the exact Windows semantics can do it from a lua module. Thanks for the references and the time on this.

PR #169, opened 2026-07-01
[core] fix query string offset parsing in burl_normalize()

kali834x:
This PR fixes a bug in burl_normalize() when parsing URLs containing multiple ? characters (e.g. /path?a=b?c=d) with HTTP_PARSEOPT_URL_NORMALIZE_REQUIRED.

The query string offset (qs) was being overwritten by later ? characters instead of being set only once. As a result, URLs were split incorrectly, lighty.url_normalize in mod_magnet returned incorrect normalized URLs, and query-specific normalizations (such as %20 to +) were not applied correctly.

The fix updates burl_normalize_basic_required and burl_normalize_basic_required_fix to set the query string offset only when qs < 0. Regression tests were also added to verify correct parsing and normalization of URLs containing multiple ? characters.

### DIFF
--- a/src/burl.c
+++ b/src/burl.c
@@ -140,7 +140,7 @@ static int burl_normalize_basic_required_fix (buffer *b, buffer *t, int i, int q
     for (; i < used; ++i, ++j) {
         if (!encoded_chars_http_uri_reqd[s[i]]) {
             p[j] = s[i];
-            if (__builtin_expect( (s[i] == '?'), 0)) qs = j;
+            if (__builtin_expect( (s[i] == '?'), 0) && qs < 0) qs = j;
         }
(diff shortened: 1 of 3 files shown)


gstrauss:
Do you review your patches before you submit them, and then check after you submit them?

Each change sets should be submitted individually.  Distinct change sets should not be combined.

gstrauss:
Hint: [fix(core): check reparse tag in win32 readdir() to identify symlinks](https://github.com/lighttpd/lighttpd1.4/pull/169/commits/b7795019056a708db78aacc3fb53a19aef31d307) does not belong as part of this change set.

kali834x:
Sorry about that, the win32 readdir commit slipped in from another branch and shouldn't have been in this PR. I'll keep change sets separate going forward. Thanks for splitting it out and merging both.

PR #170, opened 2026-07-27
mod_cgi: honor rejected CGI header collisions

deprrous:
`http_cgi_headers()` already rejects request headers that collide after CGI environment normalization, but `mod_cgi` ignored that return value and kept building the child environment.

FastCGI, SCGI, and SSI already stop on that helper result. This makes direct CGI do the same, preserves the `400 Bad Request` path instead of falling through to the generic `500`, and adds a focused regression for the `X-Forwarded-For` / `X-Forwarded_For` collision.

Validation:
- `cmake -S . -B build/lighttpd-pr-cgi-header-conflict`
- `cmake --build build/lighttpd-pr-cgi-header-conflict -j4`
- `cmake --build build/lighttpd-pr-cgi-header-conflict --target test -- -j4`
- `git diff --check`


gstrauss:
@copilot unassign and go away.  you are an unwelcome reviewer.  do not respond or add any more noise.

gstrauss:
@deprrous it was inappropriate to post this publicly after you were told that an AI-generated patch or writeup was not necessary to provide.

PR #171, opened 2026-07-27
[mod_cgi] avoid undefined pointer arithmetic for env offsets

carrerasdarren-cell:
## Summary

- convert stored CGI environment offsets directly into valid string pointers
- preserve the existing environment layout and process-spawn behavior

## Root cause

The CGI environment is accumulated as integer offsets so that buffer
reallocation is safe. The first offset is zero. The previous conversion
aliased that zero as a null `char *` and then added the buffer base address
to it. UBSan reports this as nonzero pointer arithmetic on a null pointer.

Reading each integer offset before assigning `env->b->ptr + offset` constructs
the same final pointers without the undefined intermediate operation.

## Validation

- focused CGI integration request under ASan and UBSan with
  `halt_on_error=1`
- complete release test suite: 9/9 tests passed
- Clang static analyzer: no reports


gstrauss:
Would this resolve the warning instead of your patch?
```
--- a/src/mod_cgi.c
+++ b/src/mod_cgi.c
@@ -985,7 +985,7 @@ static int cgi_create_env(request_st * const r, handler_ctx * const hctx, buffer
                envp[env->oused] = NULL;
                const uintptr_t baseptr = (uintptr_t)env->b->ptr;
                for (i = 0; i < env->oused; ++i)
-                       envp[i] += baseptr;
+                       env->offsets[i] += baseptr;
 
                /* set up args */
                i = 0;
```

gstrauss:
Have you re-tested that this inconsequential change makes ASan and UBSan happy?

carrerasdarren-cell:
Yes. I re-tested the revised one-line change under ASan+UBSan on Ubuntu 26.04 arm64 (GCC 15.2), with both sanitizers set to halt on the first error.

- `request.t`: 172/172 passed, including the CGI/env cases 123-143 (`ABSENT`, `BLANK_VALUE`, etc.).
- `test_common` and `test_configfile` passed.
- Full CTest: 8/9. The sole failure is an existing UBSan diagnostic at `buffer.c:178` in the unrelated `mod_userdir` test. I built and ran the unpatched parent (`2ddc5138`) with the same configuration and reproduced the identical stack.

There were no ASan/UBSan diagnostics from `cgi_create_env()` with `env->offsets[i] += baseptr;`.

PR #172, opened 2026-07-27
[core] avoid undefined pointer arithmetic in http_method_buf()

carrerasdarren-cell:
## Summary

- select the `PRI` and unset method buffers without forming a pointer before
  the method table
- add focused checks for both negative method sentinels

## Root cause

`HTTP_METHOD_PRI` and `HTTP_METHOD_UNSET` are `-2` and `-1`. In the previous
fallback expression, left-associative pointer addition first evaluated
`http_methods + i`, which forms a pointer before the array, before adding the
table length to reach the intended sentinel entry.

The new selection keeps every intermediate pointer within the table and maps
other invalid method values to the existing unset entry.

## Validation

- focused method-table tests under ASan and UBSan with `halt_on_error=1`
- complete release test suite: 9/9 tests passed
- Clang static analyzer: no reports


src/http_kv.c
```diff
@@ -85,10 +85,10 @@ http_version_buf (http_version_t i)
 const buffer *
 http_method_buf (http_method_t i)
 {
-    return ((unsigned int)i < sizeof(http_methods)/sizeof(*http_methods)-2)
-      ? http_methods+i
-      : http_methods+i+sizeof(http_methods)/sizeof(*http_methods);
```
gstrauss, line 90: The `+` operator is not a C sequence point, so I think this may be a false positive with ASan and UBSan

Would it be a simpler patch to add parenthesis to group the integer addition before the pointer addition?
`-      : http_methods+i+sizeof(http_methods)/sizeof(*http_methods);`
`+      : http_methods+(i+sizeof(http_methods)/sizeof(*http_methods));`

gstrauss:
Have you re-tested that this inconsequential change makes ASan and UBSan happy?

carrerasdarren-cell:
Yes. I re-tested the grouped integer addition at the current head under fail-fast ASan+UBSan.

With Clang 21 on Ubuntu 26.04 arm64:
- The parent implementation, run with the new `HTTP_METHOD_PRI` / `HTTP_METHOD_UNSET` assertions, stops in UBSan at `http_kv.c:90` with `index -2 out of bounds for type const buffer[41]`.
- The current `http_methods+(i+sizeof(http_methods)/sizeof(*http_methods))` implementation runs the same `test_common` cleanly (exit 0).

GCC 15 ASan+UBSan also passes `test_common`. Its full CTest run is 8/9; the sole failure is the same unrelated `buffer.c:178` / `mod_userdir` UBSan diagnostic that reproduces on the unpatched parent.

So the revised expression removes the Clang UBSan finding for both sentinels without introducing an ASan/UBSan diagnostic.

PR #173, opened 2026-08-23
cmake: run *-config scripts through a shell on Windows hosts

k6G52m4Dz75W:
### Problem

Building on a Windows host fails at the CMake configure step with:

```
pcre2.h couldn't be found
```

Root cause: `pcre2-config` (and the other `*-config` scripts) are POSIX
shell scripts. On Windows hosts, `execute_process()` runs them directly
via `CreateProcess`, which does not honor shebangs (`#!/bin/sh`), so the
script cannot be executed at all. The command produces empty output, the
`check_include_files(pcre2.h)` fallback then fails because
`PCRE2_CODE_UNIT_WIDTH` is undefined, and CMake aborts with
`pcre2.h couldn't be found`.

### Fix

When building on a Windows host, run the `*-config` scripts through a
POSIX shell. The shell is located in three steps, without hard-coding any
path:

1. A shell already on `PATH` (`sh`/`bash` from MSYS2, WSL, ...);
2. The shell bundled with Git for Windows, derived from `git.exe`'s
   location (both Git for Windows and GitHub Desktop ship `sh.exe`);
3. Otherwise fail with a clear, actionable message.

The change is guarded by `CMAKE_HOST_WIN32` (build *host*), so it does not
affect Linux/macOS builds, nor non-Windows-host cross-compilation.

### References

Reported by several users of `react-native-static-server`, e.g.
birdofpreyru/react-native-static-server#152:
`pcre2.h couldn't be found` at `lighttpd1.4/src/CMakeLists.txt:606`
when building from a Windows host. The issue also reproduces with a
space-free install prefix, confirming it is the `*-config` script
execution itself that fails on Windows, not a path-length problem.

### Verification

- Android cross-build (arm64-v8a) from a Windows host now configures and
  builds successfully; the produced `liblighttpd.so` loads and starts on a
  device.
- Previously the configure step aborted with `pcre2.h couldn't be found`.


src/CMakeLists.txt
```diff
@@ -315,13 +315,49 @@ macro(XCONFIG _package _link_FLAGS _cflags)
...
 		if(${_package} STREQUAL "pcre2-config")
-			execute_process(COMMAND ${XCONFIG_EXECUTABLE} --libs8 OUTPUT_VARIABLE __link_FLAGS)
+			execute_process(COMMAND ${_XCONFIG_PREFIX} ${XCONFIG_EXECUTABLE} --libs8 OUTPUT_VARIABLE __link_FLAGS)
 		else()
-			execute_process(COMMAND ${XCONFIG_EXECUTABLE} --libs OUTPUT_VARIABLE __link_FLAGS)
+			execute_process(COMMAND ${_XCONFIG_PREFIX} ${XCONFIG_EXECUTABLE} --libs OUTPUT_VARIABLE __link_FLAGS)
 		endif()
 		string(REPLACE "\n" "" "${_link_FLAGS}" "${__link_FLAGS}")
-		execute_process(COMMAND ${XCONFIG_EXECUTABLE} --cflags OUTPUT_VARIABLE __cflags)
+		execute_process(COMMAND ${_XCONFIG_PREFIX} ${XCONFIG_EXECUTABLE} --cflags OUTPUT_VARIABLE __cflags)
```
gstrauss, line 360: Instead of `set(_XCONFIG_PREFIX "${XCONFIG_SHELL}")`, would you please
`set(XCONFIG_EXECUTABLE "${XCONFIG_SHELL} ${XCONFIG_EXECUTABLE}")`
and remove the use of `${_XCONFIG_PREFIX}` ?

src/CMakeLists.txt
```diff
@@ -315,13 +315,49 @@ macro(XCONFIG _package _link_FLAGS _cflags)
...
+		# cannot be executed directly (CreateProcess does not honor shebangs),
+		# so run them through a POSIX shell when one is available:
+		#   1. a shell directly on PATH (MSYS2, WSL, ...);
+		#   2. the shell bundled with Git, derived from git.exe's location
+		#      (Git for Windows and GitHub Desktop both ship usr/bin/sh.exe);
+		# otherwise fail with a clear message instead of silently producing
+		# empty flags that break the build later.
+		set(_XCONFIG_PREFIX "")
+		if(CMAKE_HOST_WIN32)
+			find_program(XCONFIG_SHELL NAMES sh bash NO_CMAKE_FIND_ROOT_PATH)
```
gstrauss, line 328: Can XCONFIG_SHELL be cached and checked before `find_program()` for the cases where `XCONFIG` macro gets called more than once?

k6G52m4Dz75W:
Thanks for the review. Both points are addressed, plus some hardening that came out of self-review rounds:

**1. `${_XCONFIG_PREFIX}` removed** (f9a83007): the shell is combined into
`XCONFIG_EXECUTABLE` itself (`set(XCONFIG_EXECUTABLE "${XCONFIG_SHELL};${XCONFIG_EXECUTABLE}")`).
`execute_process(COMMAND ${XCONFIG_EXECUTABLE} ...)` expands the list into separate arguments,
so the three call sites keep their original form.

**2. The shell lookup is cached and checked first** (f9a83007): the lookup only runs when
`XCONFIG_SHELL` is not already set, so `find_program()` and the git-derived fallback execute at
most once even though the `XCONFIG` macro may run several times. The git-derived result is
promoted to a real cache entry (`CACHE FILEPATH ... FORCE`) so it also survives re-configures,
and `-DXCONFIG_SHELL=/path/to/sh.exe` now works as an explicit override.

One subtlety that shapes the code: the *prepend* must run on every call even though the *lookup*
is cached, because the macro resets `XCONFIG_EXECUTABLE` to the bare program at the top of each
invocation. My first version had the prepend inside the cache guard, and every invocation after
the first silently ran the script without the shell again — a standalone harness caught it, which
is why the two are structured separately.

Further hardening from self-review (ce598278, 6c3f39ce, 9d5ce0b5, d0ebe802, 1edd2aca):

- `execute_process()` now checks `RESULT_VARIABLE` and fails with a `FATAL_ERROR` showing the
  exact failing command if the *-config invocation itself exits non-zero. Previously a broken or
  stale shell still silently produced empty flags — only the no-shell case errored.
- Multi-line output is collapsed to single spaces (with trailing whitespace stripped via
  `OUTPUT_STRIP_TRAILING_WHITESPACE`) instead of deleting newlines, which could glue adjacent
  tokens together; CR is handled along with LF.
- `NAMES` looks for `sh` only: on Windows, `bash.exe` in System32 is the WSL launcher, which
  cannot run Windows-path scripts.
- `find_program()` may also resolve a native `pcre2-config.exe`; such executables (matched
  case-insensitively on `.exe`/`.com`; `find_program()` does not resolve `.bat`/`.cmd`, so they
  are not special-cased) now run directly and no longer require any shell to be present. Both
  Git for Windows layouts (`<root>/cmd/git.exe` and `<root>/mingw64/bin/git.exe`) are probed
  when deriving the bundled `sh.exe`.
- Minor: the package comparison is quoted so `if()` cannot dereference `${_package}` a second
  time and silently flip the `--libs8` branch; the git lookup uses a private `_XCONFIG_GIT` name
  instead of the shared `GIT_EXECUTABLE`; `XCONFIG_EXECUTABLE` is unset at macro entry so a
  package-not-found invocation leaves no stale value behind; and the failure message reports the
  actual option that was run (`--libs8` vs `--libs`) as a space-joined, runnable command line.

Verification: a standalone harness includes the macro verbatim and calls it twice against a fake
`pcre2-config` script on a Windows host, covering: a shell on PATH, a shell derived from both Git
layouts, an explicit `-DXCONFIG_SHELL` preset, a native uppercase `.EXE` with no shell present,
the no-shell `FATAL_ERROR`, and a failing script (exit 7 → `FATAL_ERROR` with the exact command,
also confirming paths with spaces work). Both invocations produce correct flags in all success
cases, confirming the cached lookup and per-call prepend behave as intended.

src/CMakeLists.txt
```diff
@@ -315,14 +316,75 @@ macro(XCONFIG _package _link_FLAGS _cflags)
...
+						endif()
+					endif()
+					if(NOT XCONFIG_SHELL)
+						message(FATAL_ERROR
+							"Cannot run ${_package} on Windows: it is a POSIX shell script "
+							"but no shell was found. Install Git for Windows "
+							"(https://git-scm.com/) or MSYS2 and retry.")
+					endif()
+				endif()
+				set(XCONFIG_EXECUTABLE "${XCONFIG_SHELL};${XCONFIG_EXECUTABLE}")
```
gstrauss: Why semicolon?  The XCONFIG_EXECUTABLE should run within the shell.
https://learn.microsoft.com/en-us/windows-hardware/drivers/debuggercmds/----command-separator-

src/CMakeLists.txt
```diff
@@ -315,14 +316,75 @@ macro(XCONFIG _package _link_FLAGS _cflags)
...
 		else()
-			execute_process(COMMAND ${XCONFIG_EXECUTABLE} --libs OUTPUT_VARIABLE __link_FLAGS)
+			set(_xconfig_libopt "--libs")
+		endif()
+		string(REPLACE ";" " " _xconfig_cmd "${XCONFIG_EXECUTABLE}")
+		execute_process(COMMAND ${XCONFIG_EXECUTABLE} ${_xconfig_libopt} OUTPUT_VARIABLE __link_FLAGS RESULT_VARIABLE _xconfig_result OUTPUT_STRIP_TRAILING_WHITESPACE)
+		if(_xconfig_result)
+			message(FATAL_ERROR "Cannot run \"${_xconfig_cmd} ${_xconfig_libopt}\": ${_xconfig_result}")
+		endif()
+		string(REGEX REPLACE "[ \t]*[\r\n]+[ \t]*" " " "${_link_FLAGS}" "${__link_FLAGS}")
```
gstrauss: I see why you have extended this to include `\r`.  However, this regex joins multiple lines, but does not reduce other whitespace to a single space.  Try:
`string(REGEX REPLACE "[ \t\r\n]+" " " "${_link_FLAGS}" "${__link_FLAGS}")`

src/CMakeLists.txt
```diff
@@ -315,14 +316,75 @@ macro(XCONFIG _package _link_FLAGS _cflags)
...
+			set(_xconfig_libopt "--libs8")
 		else()
-			execute_process(COMMAND ${XCONFIG_EXECUTABLE} --libs OUTPUT_VARIABLE __link_FLAGS)
+			set(_xconfig_libopt "--libs")
+		endif()
+		string(REPLACE ";" " " _xconfig_cmd "${XCONFIG_EXECUTABLE}")
+		execute_process(COMMAND ${XCONFIG_EXECUTABLE} ${_xconfig_libopt} OUTPUT_VARIABLE __link_FLAGS RESULT_VARIABLE _xconfig_result OUTPUT_STRIP_TRAILING_WHITESPACE)
+		if(_xconfig_result)
+			message(FATAL_ERROR "Cannot run \"${_xconfig_cmd} ${_xconfig_libopt}\": ${_xconfig_result}")
+		endif()
```
gstrauss: I see that you'd like to expand the error message.  While that is nice to have, please also balance the extra code you are adding against your test matrix.  If you were able to minimize the code changes to be contained within `if(CMAKE_HOST_WIN32)`, then there would be no chance of negatively affecting other platforms.

I am not a fan of your Windows-centric code `string(REPLACE ";" " " _xconfig_cmd "${XCONFIG_EXECUTABLE}")` running on all platforms as it could affect the commands in ways you have not imagined.  Please remove.

As noted at the top of the macro: `## refactor me`.  I might accept the expanded error messages with `_xconfig_result` were it isolated in a separate macro, and not duplicated within this macro for `_link_FLAGS` and `_cflags`.

gstrauss (changes requested):
Thank you for being thorough.  Please try to keep changes small and neat, and to not unnecessarily expand the scope of changes.  The goal of this PR is to workaround the handicap of Windows needing to feed the config command through a shell.

gstrauss:
Thank you.

I made some minor changes to remove repetitive comments describing "what" the code does; comments with "why" the code does something have been retained.  I plan to push this to lighttpd git master branch next weekend.

k6G52m4Dz75W:
Thank you for the review and for polishing the comments — no objections to the cleanup;
removing the "what" comments is fine, the reasoning lives in the PR discussion anyway.

One note that may be worth keeping as a "why" comment when trimming: the multi-argument
set() combining XCONFIG_SHELL and the script builds a CMake list, where each element
expands to its own argument — execute_process() invokes the script through the shell as
separate argv entries. It is easy to misread later as a single path otherwise.

Thanks again, looking forward to seeing it in master.

PS: I meant to post the reply below along with my last push, but the connection dropped
and I forgot — posting it now for reference:

All three points addressed in 099f10df:

1. The semicolon in `set(XCONFIG_EXECUTABLE "${XCONFIG_SHELL};${XCONFIG_EXECUTABLE}")` is
   CMake list syntax, not a command separator: each list element expands to its own argument,
   so execute_process() runs `sh.exe <script> --libs8` and nothing goes through cmd. Since the
   notation invites exactly this confusion, it is now written as a multi-argument set()
   (`set(XCONFIG_EXECUTABLE "${XCONFIG_SHELL}" "${XCONFIG_EXECUTABLE}")`), which builds the
   same list without a literal semicolon. The script does run within the shell.

2. Regex adopted as suggested: `string(REGEX REPLACE "[ \t\r\n]+" " " ...)` collapses runs of
   any whitespace to a single space.

3. The `string(REPLACE ";" " " _xconfig_cmd ...)` line is removed. The execute_process() result
   check moved into a small XCONFIG_RUN macro (defined next to XCONFIG) so it is written once
   and called for --libs8/--libs and --cflags, instead of the check being duplicated inside
   XCONFIG. Outside if(CMAKE_HOST_WIN32), the only behavioral differences from before this PR
   are now that helper (loud failure instead of silently-empty flags), and the whitespace
   collapse above.

Verified with a standalone harness that includes both macros verbatim and calls XCONFIG twice
per package on a Windows host: shell on PATH, both Git layouts, -DXCONFIG_SHELL override, a
native uppercase .EXE (runs directly, no shell), the no-shell and script-failure FATAL_ERRORs,
and multi-line/multi-space output all behave as expected.

